Issue #8, opened 2021-04-16
Account for transaction cost when warning users that they are setting stake to higher than their balance


abhconcordium:
When a user attempts to add a baker, and sets the stake to higher than their current account balance, the client will give a warning that the transaction will be rejected.

However, if a user sets the stake to less than their current balance, but more than `current balance - cost of add-baker transaction`, then the add baker transaction will be rejected because the cost of the transaction is first subtracted from account balance, before the node checks that the account balance covers the amount to be staked.

This could be a potentially common occurrence if users simply set stake to their current balance hoping to stake everything.

We should warn users in this situation, (and potentially offer to reduce the amount they are trying to stake by the transaction cost?)

Issue #5, opened 2021-04-16
Inconsistent Confirmation dialogue on insufficient energy


abhconcordium:
Most confirmation dialogues have the form
`Confirm [yN]:`
where `[yN]` indicates the expected user input, as well as default behaviour.

The confirmation dialogue given when a user manually sets `--energy` on a transaction to lower than the calculated NRG cost is 
`Do you want to increase the energy allocation to 501 NRG?  `

This should be changed to be consistent with our other confirmation dialogues.

PR #10, opened 2021-04-16
Confirmation dialogue consistency and `baker add` stake warnings fix

abhconcordium:
## Purpose

Closes issues #8 #5 

On baker warning for too high stake not accounting for transaction cost
and inconsistent confirmation dialogues

## Changes

Added clarifying comments to the askConfirm function, explaining in which cases it will not add [yN]: prefix
fixed prompts sent to askConfirm in a few cases

Added calculated transaction cost in GTU to the comparison between staked amount and available balance.
Fetches the EnergyRate using a generic function, which is what is currently in use in another branch which is implementing a more general integration of NRG -> GTU price conversions. 

## Checklist

      hard-to-understand areas.


src/Concordium/Client/Runner.hs
```diff
@@ -2255,15 +2255,29 @@ processBakerCmd action baseCfgDir verbose backend =
...
+          -- General function to fetch NrgGtu rate
+          -- This functionality is going to be more generally used in an upcoming branch adding GTU prices to NRG printouts across the client.
+          -- Invoking it directly here to avoid extra work on compatability/refactoring in the future
+          getNrgGtuRate :: ClientMonad IO Types.EnergyRate
+          getNrgGtuRate = do 
+            blockSummary <- getFromJson =<< withBestBlockHash Nothing getBlockSummary
+            case blockSummary of
+              Nothing -> do
+                logError ["Failed to retrieve NRG-GTU rate from best block hash, unable to estimate GTU cost"]
+                return 0
```
abizjak, line 2278: I think probably it is better if this returns a Maybe instead of 0. This can then be handled by the consumer.

Alternatively the error message should say something to the effect "Using default exchange rate ..."

abizjak (changes requested):

src/Concordium/Client/Runner.hs
```diff
@@ -2255,15 +2255,29 @@ processBakerCmd action baseCfgDir verbose backend =
...
          let senderAddr = naAddr . esdAddress . tcEncryptedSigningData $ txCfg
          AccountInfoResult{..} <- getAccountInfoOrDie senderAddr
+         energyrate <- getNrgGtuRate
+         let gtuTransactionPrice = Types.computeCost energyrate (tcEnergy txCfg) 
          case airBaker of
            Just AccountInfoBakerResult{..} -> logFatal [[i|Account is already a baker with ID #{aibiIdentity abirAccountBakerInfo}.|]]
            Nothing -> do
-             -- TODO: this should also take into account the estimated cost for this transaction
-             if airAmount < initialStake
+             if (airAmount - gtuTransactionPrice) < initialStake
```
td202, line 2263:
```suggestion
             if airAmount - gtuTransactionPrice < initialStake
```

src/Concordium/Client/Runner.hs
```diff
@@ -2255,15 +2255,29 @@ processBakerCmd action baseCfgDir verbose backend =
...
              then
-               logFatal [[i|Account balance (#{showGtu airAmount}) is lower than the amount requested to be staked (#{showGtu initialStake}).|]]
+               logFatal [[i|Account balance (#{showGtu airAmount}) minus the cost of the transaction (#{showGtu gtuTransactionPrice}) is lower than the amount requested to be staked (#{showGtu initialStake}).|]]
```
td202, line 2265: Would it be worth allowing the user to proceed anyway, if they wish to? This would mainly be useful for testing that such a transaction does fail correctly on the node.

src/Concordium/Client/Runner.hs
abhconcordium: fixed in latest commit

src/Concordium/Client/Runner.hs
abhconcordium: Added in latest commit

src/Concordium/Client/Runner.hs
abhconcordium: I added some more information to the error message in the latest commit, specifying that the "default" behaviour is that it falls back to print all NRG-derived GTU values as 0. 
This, (now that it's more clearly communicated,) feels like the easiest/safest default behaviour, given that this function should for now exclusively be used for computing GTU values for informational prints.

If 0 isn't ideal, is there a different default value you would recommend?

Alternatives could be erroring out, or asking the user to rerun the command (assuming the edgecase of best block being dropped by the node due to finalization/branch switches in between calls.)

Maybe feels like overkill, but could be an option if we want to use it for more "mission critical" functionality later.

abizjak: I think this is Ok, but I would document this caveat about the function, especially if this is going to be a top-level function in the future.

abizjak:
@abhconcordium can you make this branch up-to-date with main so we can get it merged?

The current diff looks quite strange, it shows a bunch of things that are already on main as part of this PR.

eb-concordium (approved):

PR #14, opened 2021-04-21
Made possible to encrypt/decrypt baker keys

eb-concordium:
## Purpose

The purpose of this PR is to make it possible to encrypt baker keys when generating them and decrypt them when adding them or setting them via `baker add` and `baker set-key`.

## Changes

It is now possible to optionally encrypt the baker keys and to decrypt them if they are encrypted. 

## Checklist

      hard-to-understand areas.

### DIFF
--- a/src/Concordium/Client/Runner.hs
+++ b/src/Concordium/Client/Runner.hs
@@ -1141,8 +1141,8 @@ bakerAddTransaction baseCfg txOpts f batcBakingStake batcRestakeEarnings confirm
   bakerKeysMaybeEncrypted <- handleReadFile BS.readFile f
   let pwdAction = askPassword "Enter password for decrypting baker keys: "
   (bakerKeys, wasEncrypted) <- Password.decodeMaybeEncrypted pwdAction bakerKeysMaybeEncrypted >>= \case
-         (Left err, _) -> logFatal [printf "error: %s" err]
-         (Right v, b) -> return (v, b)
+         Left err -> logFatal [printf "error: %s" err]
+         Right (v, b) -> return (v, b)
   
   let electionSignKey = bkElectionSignKey bakerKeys
       signatureSignKey = bkSigSignKey bakerKeys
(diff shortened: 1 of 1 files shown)


src/Concordium/Client/Runner.hs
```diff
@@ -1137,8 +1137,13 @@ bakerAddTransaction :: BaseConfig -> TransactionOpts (Maybe Types.Energy)
                     -> IO (BakerKeys, TransactionConfig, Types.EncodedPayload)
 bakerAddTransaction baseCfg txOpts f batcBakingStake batcRestakeEarnings confirm = do
   encSignData <- getAccountCfgFromTxOpts baseCfg txOpts
-  bakerKeys <- eitherDecodeFileStrict f >>= getFromJson
-
+   
+  bakerKeysMaybeEncrypted <- handleReadFile BS.readFile f
+  let pwdAction = askPassword ("Enter password for decrypting baker keys:")
```
abizjak, line 1142:
```suggestion
  let pwdAction = askPassword ("Enter password for decrypting baker keys: ")
```

src/Concordium/Client/Runner.hs
```diff
@@ -2227,8 +2232,16 @@ processBakerCmd action baseCfgDir verbose backend =
   case action of
     BakerGenerateKeys outputFile -> do
       keys <- generateBakerKeys
-      let out = AE.encodePretty keys
-
+      -- let out = AE.encodePretty keys
```
abizjak, line 2235: Perhaps delete.

src/Concordium/Client/Runner.hs
```diff
@@ -2227,8 +2232,16 @@ processBakerCmd action baseCfgDir verbose backend =
...
+      pwd <- askPassword "Enter password for encryption of baker keys (leave blank for no encryption): "
+      out <- case Password.getPassword pwd of
+        "" -> return $ AE.encodePretty keys
```
abizjak, line 2238: Perhaps you could log here "empty password, not encrypting baker keys" or similar.

src/Concordium/Client/Runner.hs
```diff
@@ -2263,7 +2276,7 @@ processBakerCmd action baseCfgDir verbose backend =
              then
                logFatal [[i|Account balance (#{showGtu airAmount}) is lower than the amount requested to be staked (#{showGtu initialStake}).|]]
              else do
-               sendAndMaybeOutputCredentials bakerKeys accountKeysFile outputFile txCfg pl intOpts
+               sendAndMaybeOutputCredentials bakerKeys bakerKeysFile outputFile txCfg pl intOpts
```
abizjak, line 2279: This `sendAndMaybeOutputCredentials` still outputs keys in plaintext right? It does not ask for password like the key generation command.

abizjak (changes requested):
The `BakerGenerateKeys` command should also be updated so that it can encrypt the baker keys.

abizjak (approved):

PR #16, opened 2021-04-23
Fix print.

abizjak:
## Purpose

Fix the print command so it does not crash when updating keys.

## Checklist

      hard-to-understand areas.


tschudid (approved):
Looks good and fixes an existing problem.

PR #22, opened 2021-05-04
Error when credential holder or key index from signers option is out …

### LINKED ISSUE #6: Multi signatures does not give an error when a given key index does not exist
**Bug Description**
When specifying credential and key indexes to use to sign a transaction with the `--signers` flag, the client doesn't warn/error if one of the provided indexes is invalid, and instead silently skips that key/signature.

**Steps to Reproduce**
Try and send a transaction with the `--signers` flag set to a key that doesn't exist. For example, sending a simple transfer from account `account1`, where `account1` has a single credential with a single key. (At index 0:0)

`concordium-client.exe transaction send-gtu --sender account1 --receiver account2 --amount 1 --signers 1:1`

**Expected Result**
The client should identify that the given key does not exist and throw an error

**Actual Result**
The client fails to find the key, so just skips adding the signature to the transaction, resulting in an NRG cost 100 lower than expected, and the transaction getting sent, then rejected by the baker. (Potentially due to deserialisation failure?)

Note that if you give multiple credentials, fx `--signers 0:0,1:0,2:1`, then the client will sign with each key which it can find, and skip those it cannot find.

OS: Windows

SimonKamp:
…of bounds

## Purpose

Closes #6 

## Changes

Check if indices specified after "--signers" actually point to a key.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/src/Concordium/Client/Runner.hs
+++ b/src/Concordium/Client/Runner.hs
@@ -816,10 +816,10 @@ getAccountCfgFromTxOpts baseCfg txOpts = do
                                                                         Nothing -> False
                                                                         Just keyList -> elem k keyList) m) newKeys
       _ <- Map.traverseWithKey (\c keyIndices -> case Map.lookup c newKeys of
-                                              Nothing -> logFatal [ "No credential holder with index: " ++ (show c)]
+                                              Nothing -> logFatal [ "No credential holder with index " ++ (show c) ++ "."]
                                               Just credHolderMap -> do
                                                                   let warnIfMissingKey keyIndex = case Map.lookup keyIndex credHolderMap of
-                                                                          Nothing -> logFatal [ "No key with index: " ++ (show keyIndex) ++ " for credential holder: " ++ (show c)]
+                                                                          Nothing -> logFatal [ "No key with index " ++ (show keyIndex) ++ " for credential holder " ++ (show c) ++ "."]
                                                                           Just _ -> return() -- Key found, do nothing. 
                                                                             -- We could add the key to a map in this case, replacing the intersection and mapWithKey steps above.
                                                                   mapM_ warnIfMissingKey keyIndices) chosenKeys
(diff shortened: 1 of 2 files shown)


src/Concordium/Client/Runner.hs
```diff
@@ -815,6 +815,15 @@ getAccountCfgFromTxOpts baseCfg txOpts = do
       let filteredKeys = Map.mapWithKey (\c m -> Map.filterWithKey (\k _ -> case Map.lookup c chosenKeys of
                                                                         Nothing -> False
                                                                         Just keyList -> elem k keyList) m) newKeys
+      _ <- Map.traverseWithKey (\c keyIndices -> case Map.lookup c newKeys of
+                                              Nothing -> logFatal [ "No credential holder with index: " ++ (show c)]
+                                              Just credHolderMap -> do
+                                                                  let warnIfMissingKey keyIndex = case Map.lookup keyIndex credHolderMap of
+                                                                          Nothing -> logFatal [ "Key index: " ++ (show keyIndex) ++ " is undefined for credential holder: " ++ (show c)]
```
abizjak, line 822: Maybe the error message could be consistent with the one for missing credentials. For example
```
No key with index ... for credential holder ...
```

src/Concordium/Client/Runner.hs
```diff
@@ -815,6 +815,15 @@ getAccountCfgFromTxOpts baseCfg txOpts = do
...
+                                                                          Just _ -> return() -- Key found, do nothing. 
+                                                                            -- We could add the key to a map in this case, replacing the intersection and mapWithKey steps above.
+                                                                  _ <- mapM (warnIfMissingKey) keyIndices
+                                                                  return ()) chosenKeys
```
abizjak, line 826: These last two lines in the function can be simplified into `mapM_ warnifMissingKey keyIndices`

src/Concordium/Client/Runner.hs
```diff
@@ -819,11 +819,10 @@ getAccountCfgFromTxOpts baseCfg txOpts = do
                                               Nothing -> logFatal [ "No credential holder with index: " ++ (show c)]
                                               Just credHolderMap -> do
                                                                   let warnIfMissingKey keyIndex = case Map.lookup keyIndex credHolderMap of
-                                                                          Nothing -> logFatal [ "Key index: " ++ (show keyIndex) ++ " is undefined for credential holder: " ++ (show c)]
+                                                                          Nothing -> logFatal [ "No key with index: " ++ (show keyIndex) ++ " for credential holder: " ++ (show c)]
```
eb-concordium, line 822: Maybe without the colons?

abizjak:
Also @SimonKamp can you please add your bugfix to the changelog?

eb-concordium (approved):

abizjak (approved):

PR #58, opened 2021-08-27
Transaction memos

### LINKED ISSUE #52: Support sending transfers with memos via concordium client
**Task description**

Support sending the 3 new transaction types via concordium-client. This should be done by adding an optional "`--memo`" flag to 
- transfers
- scheduled transfers
- encrypted transfers

If the flag is present then the client will make the corresponding transfer with a memo, otherwise it will act as it did before.

To cover the different possibilities I think the following variants of the flag should be present
- `--memo` takes a string as an argument, and encodes it as a CBOR string in the memo
- `--memo-json` reads a JSON value from the supplied file and encodes it into CBOR
- `--memo-bin` takes a file as an argument and includes the contents as a CBOR byte array.

If something else is more convenient or desired then include the description of the options below in the comments.

Related
- https://github.com/Concordium/concordium-node/milestone/1

**Sub-tasks**
- [ ] Add the memo flags as described above.
- [ ] The client should check whether the new transactions are supported by the node or not using the node's API. It should fail with a meaningful error when the node does not yet support the given transactions.

### LINKED ISSUE #53: Display memos in transaction outcomes
**Task description**

Related to #52 concordium client should display memos for the 3 new transaction types when their status is queried.

This only applies to the `transaction status` command. The `raw GetTransactionStatus` will support this by default since it only displays the returned JSON.

The best user experience would be if we try to decode the value as CBOR and if it is 
- a number
- a string
- a byte array

then we display it nicely as that, and if it is something more complex we try to decode as JSON, and pretty print that if it is successful, and otherwise fall back to displaying the raw data.

eb-concordium:
## Purpose

To be able to send transactions with memos via concordium-client, and to warn the user when trying to send a transaction with a memo if protocol version is 1. 

- Closes #52
- Closes #53 

## Changes

The commands `send-gtu`, `send-gtu-encrypted` and `send-gtu-scheduled` have been extended with two memo flags:
- `--memo` for providing a memo as a string (which will be CBOR encoded on chain)
- `--memo-json` for providing a memo as json from a file (the json will be CBOR encoded on chain)
- `--memo-raw` for providing a memo as raw bytes from a file (the raw bytes will go on chain and thus **does not** necessarily  represent an CBOR encoding of anything)

## Checklist

      hard-to-understand areas.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -3,7 +3,7 @@
 ## Unreleased changes
 - The `account show` command can receive a credential registration ID instead of a name or address.
 
-## 1.1
+## 1.1.0
 - support sending the three new transaction types, i.e. TransferWithMemo, EncryptedTransferWithMemo
   and TransferWithScheduleAndMemo
 - show transfer memo in transaction status
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -9,6 +9,7 @@
 - show transfer memo in transaction status
 - show protocolVersion, genesisIndex, currentEraGenesisBlock and currentEraGenesisTime in 
   consensus status
+- this version is only compatible with node version 1.1.0 and later.
 
 ## 1.0.1
 
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -3,6 +3,13 @@
 ## Unreleased changes
 - The `account show` command can receive a credential registration ID instead of a name or address.
 
+## 1.1
+- support sending the three new transaction types, i.e. TransferWithMemo, EncryptedTransferWithMemo
+  and TransferWithScheduleAndMemo
+- show transfer memo in transaction status
+- show protocolVersion, genesisIndex, currentEraGenesisBlock and currentEraGenesisTime in 
+  consensus status
+
 ## 1.0.1
 
 - support the new mobile wallet export format. This breaks compatibility with
--- a/concordium-client.cabal
+++ b/concordium-client.cabal
@@ -5,7 +5,7 @@ cabal-version: 2.0
 -- see: https://github.com/sol/hpack
 
 name:           concordium-client
-version:        1.1
+version:        1.1.0
 description:    Please see the README on GitHub at <https://github.com/Concordium/concordium-client#readme>
 homepage:       https://github.com/Concordium/concordium-client#readme
 bug-reports:    https://github.com/Concordium/concordium-client/issues
--- a/concordium-client.cabal
+++ b/concordium-client.cabal
@@ -1,10 +1,10 @@
 cabal-version: 2.0
 
--- This file has been generated from package.yaml by hpack version 0.34.4.
+-- This file has been generated from package.yaml by hpack version 0.33.0.
 --
 -- see: https://github.com/sol/hpack
 --
--- hash: e81ac8f8ccfb533313a49381417d77b6ea979e40b403d3968c98c39aac04a48f
+-- hash: ac4ef84a42a671516a8c239ff0ce630e628b542f933d0454683b63cc5ddb0758
 
 name:           concordium-client
 version:        1.0.1
@@ -69,14 +69,7 @@ library
       Proto.ConcordiumP2pRpc Proto.ConcordiumP2pRpc_Fields
   hs-source-dirs:
       src
-  default-extensions:
-      OverloadedStrings
-      RecordWildCards
-      GeneralizedNewtypeDeriving
-      ScopedTypeVariables
-      FlexibleContexts
-      LambdaCase
-      TupleSections
+  default-extensions: OverloadedStrings RecordWildCards GeneralizedNewtypeDeriving ScopedTypeVariables FlexibleContexts LambdaCase TupleSections
   ghc-options: -Wall -Wcompat -Werror=missing-fields -Wredundant-constraints -O
   build-depends:
       aeson
@@ -89,6 +82,8 @@ library
     , base16-bytestring
     , base64-bytestring
     , bytestring
+    , cborg
+    , cborg-json
     , cereal
     , concordium-base
     , concurrent-extra >=0.7
@@ -126,14 +121,7 @@ executable concordium-client
       Paths_concordium_client
   hs-source-dirs:
       app
-  default-extensions:
-      OverloadedStrings
-      RecordWildCards
-      GeneralizedNewtypeDeriving
-      ScopedTypeVariables
-      FlexibleContexts
-      LambdaCase
-      TupleSections
+  default-extensions: OverloadedStrings RecordWildCards GeneralizedNewtypeDeriving ScopedTypeVariables FlexibleContexts LambdaCase TupleSections
   ghc-options: -Wall -Wcompat -Werror=missing-fields -Wredundant-constraints -O
   build-depends:
       base
@@ -154,14 +142,7 @@ executable middleware
       Paths_concordium_client
   hs-source-dirs:
       middleware
-  default-extensions:
-      OverloadedStrings
-      RecordWildCards
-      GeneralizedNewtypeDeriving
-      ScopedTypeVariables
-      FlexibleContexts
-      LambdaCase
-      TupleSections
+  default-extensions: OverloadedStrings RecordWildCards GeneralizedNewtypeDeriving ScopedTypeVariables FlexibleContexts LambdaCase TupleSections
   ghc-options: -Wall -threaded
   build-depends:
       aeson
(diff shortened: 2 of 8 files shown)


concordium-client.cabal
```diff
@@ -1,10 +1,10 @@
 cabal-version: 2.0
 
--- This file has been generated from package.yaml by hpack version 0.34.4.
+-- This file has been generated from package.yaml by hpack version 0.33.0.
```
abizjak, line 3: Please update your stack so that it does not regenerate these files and introduce this diff. Do `stack upgrade`.

src/Concordium/Client/Commands.hs
```diff
@@ -21,11 +21,12 @@ module Concordium.Client.Commands
   , BakerCmd(..)
   , IdentityCmd(..)
   , IdentityShowCmd(..)
+  , MemoInput(..)
   ) where
 
 import Data.Text hiding (map, unlines)
 import Data.Version (showVersion)
-import Data.Word (Word64)
+import Data.Word (Word64, Word8)
```
abizjak, line 29: Remove the unused import.

src/Concordium/Client/Cli.hs
```diff
@@ -390,6 +394,10 @@ instance AE.FromJSON ConsensusStatusResult where
     csrLastFinalizedTime <- v .: "lastFinalizedTime"
     csrFinalizationPeriodEMA <- v .: "finalizationPeriodEMA"
     csrFinalizationPeriodEMSD <- v .: "finalizationPeriodEMSD"
+    csrProtocolVersion <- v .:? "protocolVersion" .!= P1
+    csrGenesisIndex <- v .:? "genesisIndex" .!= 0
+    csrCurrentEraGenesisBlock <- v .:? "currentEraGenesisBlock" .!= csrGenesisBlock
+    csrCurrentEraGenesisTime <- v .:? "currentEraGenesisTime" .!= csrGenesisTime
```
abizjak, line 400: It would be good to comment why these are sensible defaults.

src/Concordium/Client/Runner.hs
```diff
@@ -116,6 +117,11 @@ import           Text.Printf
 import           Text.Read (readMaybe)
 import Data.Time.Clock (addUTCTime, getCurrentTime, UTCTime)
 import Data.Ratio
+import Codec.CBOR.Write
+import Codec.CBOR.Encoding
+import Codec.CBOR.JSON
+import Data.ByteString (ByteString)
+import qualified Data.ByteString.Lazy.Char8 as BSL
```
abizjak, line 124: Remove unused import.

src/Concordium/Client/Runner.hs
```diff
@@ -500,6 +506,24 @@ loadAccountImportFile format file name = do
      accCfg <- decodeGenesisFormattedAccountExport contents name pwd `withLogFatalIO` ("cannot import account: " ++)
      return [accCfg]
 
+-- Read a transaction memo as either
+-- - a CBOR encoded string,
+-- - CBOR encoded json (from a file), or
+-- - raw bytes (from a file). 
+readMemoAsBytestring :: MemoInput -> Backend -> IO ByteString
+readMemoAsBytestring memo backend = do
```
abizjak, line 514: I think the following function should either take the protocol version as a parameter so it does not have to use the backend itself, or it should be changed so that it lives in the `ClientMonad IO`.

The reason for the latter is that as it is now whenever this function is used it creates a connection to the node, and then drops it. The connection is later re-established to send the transaction.

Since this function also checks whether a memo is allowed it should probably be renamed to reflect this, something along the lines of `checkAndGetMemo`.

src/Concordium/Client/Runner.hs
```diff
@@ -537,15 +561,22 @@ processTransactionCmd action baseCfgDir verbose backend =
...
 
-      let pl = Types.encodePayload $ Types.Transfer (naAddr receiverAddress) amount
+      let plDo = do
+            res <- case maybeMemo of 
+              Nothing -> return $ Types.Transfer (naAddr receiverAddress) amount
+              Just memo -> do
+                bs <- readMemoAsBytestring memo backend
+                return $ Types.TransferWithMemo (naAddr receiverAddress) (Types.Memo $ BSS.toShort bs) amount 
+            return $ Types.encodePayload res
+      pl <- plDo
```
abizjak, line 579: You do not have to introduce `plDo` (if you do not want to), e.g., you can write

```haskell
pl <- do
   res <- case maybeMemo of
      ...
   return $ Types.encodePayload res
```

If you change the `readMemoAsBytes` to live in the ClientMonad then you will need to restructure this a little bit by doing

```haskell
withClient backend $ do
   let plDo = do
       ...
   pl <- plDo
   let nrgCost = ...
   ...
   sendAndTailTransaction_ txCfg pl intOpts
```

src/Concordium/Client/Runner.hs
```diff
@@ -627,7 +664,14 @@ processTransactionCmd action baseCfgDir verbose backend =
...
 
-        let payload = Types.encodePayload $ Types.EncryptedAmountTransfer (naAddr receiverAcc) transferData
+        let plDo = do
+                res <- case maybeMemo of 
+                  Nothing -> return $ Types.EncryptedAmountTransfer (naAddr receiverAcc) transferData
+                  Just memo -> do
+                    bs <- readMemoAsBytestring memo backend
+                    return $ Types.EncryptedAmountTransferWithMemo (naAddr receiverAcc) (Types.Memo $ BSS.toShort bs) transferData
+                return $ Types.encodePayload res
+        payload <- liftIO plDo
```
abizjak, line 674: If you move `readMemoAsBytes` to `ClientMonad` then you don't need this `liftIO`.

abizjak (changes requested):
Overall it looks good, just minor code style improvements and some optimizations.

However please update the version number to 1.1 and also update the changelog to add the new functionality.

package.yaml
```diff
@@ -1,5 +1,5 @@
 name:                concordium-client
-version:             1.0.1
+version:             1.1
```
abizjak, line 2: I think we should keep the major.minor.patch versioning, so please use `1.1.0`.

src/Concordium/Client/Cli.hs
```diff
@@ -360,7 +360,25 @@ data ConsensusStatusResult = ConsensusStatusResult
...
+-- * P1 for protocolVersion 
+-- * 0 for genesisIndex since the first protocol update to increase the genesisIndex is
+--   assumed to change the protocol version to 2, and thus we can assume that the
+--   genesisIndex is 0, if it isn't returned by the node.
+-- * csrGenesisBlock for currentEraGenesisBlock since the first protocol update
+--   changes the protocol version to 2, meaning that if currentEraGenesisBlock
+--   is not returned by the node, then the current genesis block is the very
+--   first genesis block (since no regenesis would have happened yet)
+-- * csrGenesisTime for currentEraGenesisTime since (again) no regenesis 
+--   would have happened if currentEraGenesisTime is not returned by the node.
```
abizjak, line 381: We have discussed today that there is no need to keep this backwards compatibility so I recommend removing this default handling.

This will also make it easier to refactor in the future and use the ConsensusStatus type from base without it being a breaking change.

ChangeLog.md
```diff
@@ -3,6 +3,13 @@
 ## Unreleased changes
 - The `account show` command can receive a credential registration ID instead of a name or address.
 
+## 1.1
```
abizjak, line 6: Please use `1.1.0` to conform to semantic versioning (also see comment below).

abizjak (changes requested):

abizjak:
Also, please update the submodule reference to concordium-base after that is merged to main.

abizjak:
Also, please update the submodule reference to concordium-base after that is merged to main.

ChangeLog.md
```diff
@@ -3,6 +3,13 @@
 ## Unreleased changes
 - The `account show` command can receive a credential registration ID instead of a name or address.
 
+## 1.1.0
+- support sending the three new transaction types, i.e. TransferWithMemo, EncryptedTransferWithMemo
+  and TransferWithScheduleAndMemo
+- show transfer memo in transaction status
+- show protocolVersion, genesisIndex, currentEraGenesisBlock and currentEraGenesisTime in 
+  consensus status
```
abizjak, line 11:
```suggestion
  consensus status
- this version is only compatible with node version 1.1.0 and later.
```

PR #63, opened 2021-09-03
Update `GetBlocksAtHeight` query to support genesis-relative queries.

td202:
## Purpose

Support the changes to the `GetBlocksAtHeight` query in the client. (See https://github.com/Concordium/concordium-node/pull/148)

## Changes

The `raw GetBlocksAtHeight` command can take an optional genesis index, given by the `--genesis-index <GENINDEX>` option.  If supplied, the height is considered relative to the genesis block at this index. Additionally, the `--restrict` flag can be specified to only return blocks that belong to the specified genesis index (or index 0, if the genesis index was not  specified).

## Checklist

      hard-to-understand areas.


abizjak (commented):
The changes are OK, but I think we should also add the additional fields to the output of `block show` (i.e. in printBlockInfo) as part of this PR, since those were also added to the RPC interface.

eb-concordium (approved):

abizjak (approved):

PR #65, opened 2021-09-07
Fix static build by using integer-gmp variant of the compiler

abizjak:
## Purpose

cborg is not compatible with integer-simple and 8.10.4. 

## Changes

The cborg dependency did not have the flag to disable integer-gmp dependency. concordium-client now depends on our own fork that adds this flag.

## Checklist

      hard-to-understand areas.


abizjak:
I have verified that all of these work now.

scripts/distributables/macos-concordium-client.Jenkinsfile
```diff
@@ -29,26 +30,16 @@ pipeline {
...
 
-                    ghc-variant: integer-simple
-                    EOF
-                    
+                    # due to a bug in stack and flag propagation we have to do this hack
                     sed -i '' "s/default: False/default: True/g" deps/concordium-base/package.yaml
+                    sed -i '' "s/default: False/default: True/g" deps/concordium-base/concordium-base.cabal
+
+                    # At present concordium-client does not build in a static build with any version of ghc more recent than 8.8.4.
+                    stack build --compiler=ghc-8.8.4 --flag concordium-client:-middleware
```
bisgardo, line 41: Maybe use `GHC_VERSION` instead of hardcoding?

bisgardo (approved):

scripts/distributables/macos-concordium-client.Jenkinsfile
abizjak: The thing is that it is deliberately hardcoded since it does not work with other versions.

Issue #34, opened 2021-07-06
Show smart contract state as bytes, when schema does not include type of state


limemloh:
**Task description**
Currently, when running `concordium-clien contract show ...` the state is only displayed as Null, when there is a schema present, which does not include a schema type for the state.

The expected behavior would be to let the user know, no schema type for the state was found, and then show the state as bytes.

PR #66, opened 2021-09-07
Show raw state if schema type not found for state

eb-concordium:
## Purpose

To show smart contract state as bytes, when schema does not include type of state. Closes https://github.com/Concordium/concordium-client/issues/34.

## Changes

* The `WithSchema` constructor of the type `ContractInfo` has been extended to contain the raw bytes, so that these can be printed `printContractInfo` in the case, where a schema is provided but doesn't include the state type.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/src/Concordium/Client/Types/Contract/Info.hs
+++ b/src/Concordium/Client/Types/Contract/Info.hs
@@ -11,7 +11,7 @@ import qualified Concordium.Client.Types.Contract.Schema as CS
 import qualified Concordium.Client.Types.Contract.Parameter as CP
 import qualified Concordium.Wasm as Wasm
 
-import Data.Aeson ((.=), (.:))
+import Data.Aeson ((.:))
 import qualified Data.Aeson as AE
 import Data.ByteString (ByteString)
 import qualified Data.ByteString as BS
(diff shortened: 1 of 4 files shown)


src/Concordium/Client/Types/Contract/Info.hs
```diff
@@ -73,8 +68,16 @@ data ContractInfo
 
 -- TODO: Use CS.ContractSchema to avoid having to lookup by contract name.
 data ContractState =
-    WithSchema !CS.ModuleSchema !AE.Value  -- ^ The module schema and the decoded contract state.
-  | JustBytes !ByteString                  -- ^ The binary-encoded contract state.
+    WithSchema
+    {
+      wsModuleSchema :: !CS.ModuleSchema, -- ^ The module schema
+      wsValue :: !(Maybe AE.Value), -- ^ The decoded contract state, if schema type found for the state.
```
abizjak, line 74: I'm not sure I understand the idea here. Why is this a maybe? Why does the other case not apply in this case.

If this has to be like it is then please provide a comment as to why. The current comment would indicate that if this is Nothing then the `JustBytes` case would apply.

abizjak (approved):

PR #88, opened 2021-10-18
Inspect modules without schemas

### LINKED ISSUE #44: Retrieve init and receive methods from a module on chain
Concordium-client currently does not retrieve these, so it cannot check that the supplied method is correct.

There are two things here. One is it can use the schema for this, but it can also parse the export list and check directly without the schema.

The main difficulty is a parser for the export section of Wasm, but that again is pretty straightforward.

Bargsteen:
## Purpose

Improves `module inspect` so that:
-  It will show *all* contracts from a module irrespective of whether a schema is included or not.
- It will show receive methods as well.

Closes #44

## Changes

- Change the parsing of wasm modules so both schemas and exported function names are found
- Print all contracts and receive functions in a module including optional schemas
- Warn the user if the schema has signatures for methods not in the module
- Add some helper functions and refactor a bit

## Checklist

      hard-to-understand areas.

### DIFF
--- a/src/Concordium/Client/Output.hs
+++ b/src/Concordium/Client/Output.hs
@@ -383,41 +383,63 @@ printModuleInspectInfo namedModRef moduleSchema exportedFuncNames = do
             mkMapFromInits :: [FuncName] -> Map.Map Text ContractSigs
             mkMapFromInits = Map.fromList . foldr mkInitTuples []
               where mkInitTuples x xs = case x of
-                      InitFuncName name -> (name, ContractSigs { csInitSig = Nothing, csReceiveSigs = mempty }) : xs
+                      InitFuncName name -> (name, ContractSigs { csInitSig = Nothing, csReceiveSigs = Map.empty }) : xs
                       ReceiveFuncName _ _ -> xs
 
             insertReceiveNames :: [FuncName] -> Map.Map Text ContractSigs -> Map.Map Text ContractSigs
-            insertReceiveNames [] m = m
-            insertReceiveNames (InitFuncName _:remaining) m = insertReceiveNames remaining m
-            insertReceiveNames (ReceiveFuncName cname fname:remaining) m = case Map.lookup cname m of
-                                                                            Nothing -> insertReceiveNames remaining m -- This should never happen, as we validate modules before they are put on chain.
-                                                                            Just ContractSigs {..} -> let updatedCsReceiveSigs = Map.insert fname Nothing csReceiveSigs
-                                                                                                          m' = Map.insert cname (ContractSigs {csInitSig = csInitSig, csReceiveSigs = updatedCsReceiveSigs}) m
-                                                                                                       in insertReceiveNames remaining m'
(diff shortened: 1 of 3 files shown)


src/Concordium/Client/Output.hs
```diff
@@ -340,17 +340,34 @@ printContractInfo CI.ContractInfo{..} namedOwner namedModRef = do
...
-    contracts = map showContract $ Map.toList contractSchemas
-    showContract (contractName, CS.ContractSchema{..}) = case initSig of
+    contractsFromSchema :: Map.Map Text (Maybe SchemaType)
+    contractsFromSchema = case moduleSchema of
+      Nothing -> Map.empty
+      Just CS.ModuleSchema{..} -> fmap initSig contractSchemas
+
+    -- Find the contract names, strip the `init_` prefix, and create a Map without any corresponding init signatures.
+    contractsFromExports :: Map.Map Text (Maybe SchemaType)
+    contractsFromExports = Map.fromList . map ((, Nothing) . Text.drop 5) . filter (Text.isPrefixOf "init_") $ exportedFuncNames
```
abizjak, line 363: I think we have a function 'isValidInitName' which should be used instead of `isPrefixOf` I think.

src/Concordium/Client/Output.hs
```diff
@@ -340,17 +340,34 @@ printContractInfo CI.ContractInfo{..} namedOwner namedModRef = do
...
+
+    -- Combine the contracts from the schema and the exports, preferring the values from the schema map.
+    combinedContracts :: Map.Map Text (Maybe SchemaType)
+    combinedContracts = Map.union contractsFromSchema contractsFromExports
```
abizjak, line 367: If the schema has functions that do not exist in the module then they should not be included, right?

In such a case I think we need to signal an error or a warning that something is off.

src/Concordium/Client/Types/Contract/Schema.hs
```diff
@@ -1,6 +1,18 @@
...
+                                               ModuleSchema(..),
+                                               SchemaType(..),
+                                               SizeLength(..),
+                                               decodeEmbeddedSchema,
+                                               decodeEmbeddedSchemaAndExports,
+                                               decodeModuleSchema,
+                                               getListOfWithKnownLen,
+                                               getListOfWithSizeLen,
+                                               lookupSignatureForFunc,
+                                               putLenWithSizeLen) where
```
abizjak, line 15: Is this really the best possible indentation?

I would recommend not having all this whitespace on the left.

src/Concordium/Client/Output.hs
```diff
@@ -340,17 +340,34 @@ printContractInfo CI.ContractInfo{..} namedOwner namedModRef = do
...
+
+    contracts = map showContract $ Map.toList combinedContracts
+    showContract (contractName, initSig) = case initSig of
```
abizjak, line 370: Shouldn't we also print the methods for contracts (aka entrypoints)? We have the information available.

PR #89, opened 2021-10-18
Clarify info message for when default account is used

### LINKED ISSUE #9: Inconsistent/confusing behaviour between commands when no --sender account is specified
For most commands, if a required parameter is missing, this will be clearly communicated to the user. For example, a `transaction send-gtu` command with no `--receiver` parameter will print

```
Missing: --receiver RECEIVER-ACCOUNT

Usage: concordium-client.exe transaction send-gtu --receiver RECEIVER-ACCOUNT --amount GTU-AMOUNT
                                                  [--sender SENDER] [--keys KEYS] [--signers SIGNERS]
                                                  [--nonce NONCE] [--energy MAX-ENERGY] [--expiry EXPIRY]
                                                  [--no-confirm] [--no-wait]
  Transfer GTU from one account to another.
```

However for some commands, if the `--sender` parameter is missing, it will instead try and use a default account, and if this default account is not found, it will throw an error

```
Account reference not provided; using "default".
Error: The identifier 'default' is neither the address nor the name of an account.
```

This could be made more informative by also informing the user that the `--sender` parameter is not set.

Bargsteen:
## Purpose

Clarify the info and error messages written when the `default` account is used. 
Closes #9.

## Changes

- Clarify the info message printer to include the argument/flag omission.
- Provide a better error message when the argument/flag was omitted, but no default account exists (this was particularly confusing).
- I've tried to phrase the info and error messages so that the default-account feature can be discovered by accident.
- Remove an unneeded `Maybe`. 

## Checklist

      hard-to-understand areas.

### DIFF
--- a/src/Concordium/Client/Config.hs
+++ b/src/Concordium/Client/Config.hs
@@ -843,11 +843,13 @@ getAccountConfig :: Maybe Text
 getAccountConfig account baseCfg keysDir keyMap encKey autoInit = do
   account' <- case account of
     Nothing -> do
-      logInfo [[i|"the --sender flag is not provided; using the default account name '#{defaultAccountName}'"|]]
+      logInfo [[i|the --sender flag is not provided; using the default account name '#{defaultAccountName}'|]]
       return defaultAccountName
     Just a -> return a
   namedAddr <- case getAccountAddress (bcAccountNameMap baseCfg) account' of
-                 Left err -> logFatal [err]
+                 Left err -> if isNothing account
+                             then logFatal [[i|The --sender flag was not provided; so the default account name '#{defaultAccountName}' was used, but no account with that name exists.|]]
+                             else logFatal [err]
                  Right v -> return v
 
   (bc, km, cidMap) <-
--- a/src/Concordium/Client/Runner.hs
+++ b/src/Concordium/Client/Runner.hs
@@ -1543,7 +1543,7 @@ processAccountCmd action baseCfgDir verbose backend =
 
       input <- case inputMaybe of
         Nothing -> do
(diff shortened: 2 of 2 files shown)


src/Concordium/Client/Runner.hs
```diff
@@ -170,10 +170,10 @@ getFromJsonAndHandleError handleError r = do
 
 -- |Look up account from the provided name or address. If 'Nothing' is given, use the defaultAcccountName.
 -- Fail if the address cannot be found.
--- Should be kept consistent with processAccountCmd for AccountShow
 getAccountAddressArg :: AccountNameMap -> Maybe Text -> IO NamedAddress
-getAccountAddressArg m input =
-  case getAccountAddress m $ fromMaybe defaultAccountName input of
+getAccountAddressArg m input = do
```
abizjak, line 174: As far as I can tell this function is only ever used with the second argument being `Just _`. So it can be simplified.

src/Concordium/Client/Config.hs
```diff
@@ -841,11 +841,7 @@ getAccountConfig :: Maybe Text
...
                  -> IO (BaseConfig, AccountConfig)
 getAccountConfig account baseCfg keysDir keyMap encKey autoInit = do
-  account' <- case account of
-    Nothing -> do
-      logInfo [printf "account reference not provided; using \"%s\"" defaultAccountName]
-      return defaultAccountName
-    Just a -> return a
+  account' <- getAccountOrDefault account
   namedAddr <- case getAccountAddress (bcAccountNameMap baseCfg) account' of
                  Left err -> logFatal [err]
```
abizjak, line 846: Perhaps it would be useful to revise the error message here in case we attempted to use the "default" account.
We could make it less confusing.

PR #105, opened 2021-11-30
Remove the threaded option from tx-generator

simonkamp:
## Purpose

The tx-generator tool deadlocks on windows when compiled with -threaded

## Changes

Removed the threaded flag.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/package.yaml
+++ b/package.yaml
@@ -133,7 +133,6 @@ executables:
       - -Werror=missing-fields
       - -Wredundant-constraints
       - -O
-      - -threaded
 
     when:
       - condition: flag(static)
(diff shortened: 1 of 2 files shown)


package.yaml
```diff
@@ -133,7 +133,6 @@ executables:
       - -Werror=missing-fields
       - -Wredundant-constraints
       - -O
-      - -threaded
```
abizjak, line 136: Please only do this for the broken platform (Windows in this case).

And please add a comment why it is needed so somebody does not introduce it later.

td202 (approved):

abizjak (approved):

PR #115, opened 2022-03-09
Adds calling fallback entry point feature to `concordium-client`

rimbi:
## Purpose

When calling `contract update` with an invalid entrypoint
the fallback entrypoint is called if one specified in the contract.

The name of the fallback entrypoint is assumed to be an empty string.

## Changes

Checks the existence of specified and default entrypoint names.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/src/Concordium/Client/Output.hs
+++ b/src/Concordium/Client/Output.hs
@@ -634,12 +634,12 @@ showRejectReason verbose = \case
       "account does not exist"
   Types.InvalidInitMethod m name ->
     if verbose then
-      printf "invalid init method '%s' of module '%s'" (show name) (show m)
+      [i|invalid init method '#{name}' of module '#{m}'|]
     else
       "invalid init method"
   Types.InvalidReceiveMethod m name ->
     if verbose then
-      printf "invalid receive method '%s' of module '%s'" (show name) (show m)
+      [i|invalid receive method '#{name}' of module '#{m}'|]
     else
       "invalid receive method"
   Types.InvalidModuleReference m ->
--- a/src/Concordium/Client/Runner.hs
+++ b/src/Concordium/Client/Runner.hs
@@ -1848,6 +1848,31 @@ data ModuleDeployTransactionCfg =
 moduleDeployTransactionPayload :: ModuleDeployTransactionCfg -> Types.Payload
 moduleDeployTransactionPayload ModuleDeployTransactionCfg {..} = Types.DeployModule mdtcModule
 
+-- |Checks if the he given receive name is valid and if so, returns it back
+-- or otherwise a fallback receive name for v1 contracts.
+checkAndGetContractReceiveName :: CI.ContractInfo -> Text -> IO Text
+checkAndGetContractReceiveName contrInfo receiveName = do
+  if CI.hasReceiveMethod receiveName contrInfo
+    then return receiveName
+    else
+      if not $ CI.hasFallbackReceiveSupport contrInfo
+        then confirmAbortOrContinue False
+        else let (_, fallbackReceiveName) = Wasm.contractAndFunctionName $ Wasm.makeFallbackReceiveName $ Wasm.ReceiveName (CI.getContractName contrInfo <> receiveName) in
+          if CI.hasReceiveMethod fallbackReceiveName contrInfo
+          then do
+            logInfo [[i|The contract does not have an entrypoint named '#{receiveName}'. The fallback entrypoint will be used instead.|]]
+            return fallbackReceiveName
+          else confirmAbortOrContinue True
+  where
+    confirmAbortOrContinue hasFallbackReceiveSupport = do
+      let fallback_msg :: String = if hasFallbackReceiveSupport
+                then " or a fallback entrypoint"
+                else ""
+      logWarn [[i|The contract does not have an entrypoint named '#{receiveName}'#{fallback_msg}.|]]
+      confirmed <- askConfirmation $ Just "Do you still want to continue? "
+      unless confirmed $ logFatal ["aborting..."]
+      return receiveName
+
 -- |Process a 'contract ...' command.
 processContractCmd :: ContractCmd -> Maybe FilePath -> Verbose -> Backend -> IO ()
 processContractCmd action baseCfgDir verbose backend =
@@ -1951,15 +1976,15 @@ processContractCmd action baseCfgDir verbose backend =
         Just (InvokerAccount nameOrAddr) -> Just . Types.AddressAccount . naAddr <$> getAccountAddressArg (bcAccountNameMap baseCfg) nameOrAddr
         Just InvokerContract{..} -> Just . Types.AddressContract . ncaAddr <$> getNamedContractAddress (bcContractNameMap baseCfg) icIndexOrName icSubindex
 
-      contrInfo <- withClient backend . withBestBlockHash block $ \bb -> getContractInfo namedContrAddr bb
+      contrInfo <- withClient backend . withBestBlockHash block $ getContractInfo namedContrAddr
       let namedModRef = NamedModuleRef {nmrRef = CI.ciSourceModule contrInfo, nmrNames = []} -- Skip finding nmrNames, as they won't be shown.
 
       let contractName = CI.getContractName contrInfo
       let wasmReceiveName = Wasm.ReceiveName [i|#{contractName}.#{receiveName}|]
-      unless (CI.hasReceiveMethod receiveName contrInfo) $ logFatal [[i|The contract does not have a receive function named '#{receiveName}'.|]]
+      updatedReceiveName <- checkAndGetContractReceiveName contrInfo receiveName
 
       modSchema <- withClient backend . withBestBlockHash block $ getSchemaFromFileOrModule schemaFile namedModRef
-      wasmParameter <- getWasmParameter parameterFile modSchema (CS.ReceiveFuncName contractName receiveName)
+      wasmParameter <- getWasmParameter parameterFile modSchema (CS.ReceiveFuncName contractName updatedReceiveName)
       let nrg = fromMaybe (Types.Energy 10_000_000) energy
 
       let invokeContext = InvokeContract.ContractContext
@@ -1981,7 +2006,7 @@ processContractCmd action baseCfgDir verbose backend =
           Right jsonValue -> case AE.fromJSON jsonValue of
             AE.Error jsonErr -> logFatal [[i|Invocation failed with error: #{jsonErr}|]]
             AE.Success InvokeContract.Failure{..} -> do
-              returnValueMsg <- mkReturnValueMsg rcrReturnValue schemaFile modSchema contractName receiveName
+              returnValueMsg <- mkReturnValueMsg rcrReturnValue schemaFile modSchema contractName updatedReceiveName
               -- Logs in cyan to indicate that the invocation returned with a failure.
               -- This might be what you expected from the contract, so logWarn or logFatal should not be used.
               log Info (Just ANSI.Cyan) [[iii|Invocation resulted in failure:\n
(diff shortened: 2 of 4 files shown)


src/Concordium/Client/Runner.hs
```diff
@@ -1851,6 +1851,21 @@ data ModuleDeployTransactionCfg =
...
 
+-- |Checks if the contract has the given receive name or otherwise a fallback receive for
+-- v1 cobtracts. Returns the receive name when applicable. 
+checkAndGetContractReceiveName :: CI.ContractInfo -> Text -> IO Text
+checkAndGetContractReceiveName contrInfo receiveName = do
+  if CI.hasReceiveMethod receiveName contrInfo
+    then return (receiveName)
+    else
+      if not $ CI.hasFallbackReceiveSupport contrInfo
+        then logFatal [[i|The contract does not have an entrypoint named '#{receiveName}'.|]]
```
abizjak, line 1862: I think this should be `logWarn` with a confirmation prompt.

We usually do this in concordium client to allow for testing, or if the user really wants to. So we prompt the user, telling them that what they want to do is almost certainly going to lead to failure, but allowing them to continue.

src/Concordium/Client/Runner.hs
```diff
@@ -1851,6 +1851,21 @@ data ModuleDeployTransactionCfg =
...
+        else let fallbackReceiveName = "" in
```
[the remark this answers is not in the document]
abizjak: That function is in concordium-base, in the new-contract-state branch which is not yet merged to main.

src/Concordium/Client/Runner.hs
```diff
@@ -1954,15 +1969,16 @@ processContractCmd action baseCfgDir verbose backend =
...
 
-      contrInfo <- withClient backend . withBestBlockHash block $ \bb -> getContractInfo namedContrAddr bb
+      contrInfo <- withClient backend . withBestBlockHash block $ getContractInfo namedContrAddr
       let namedModRef = NamedModuleRef {nmrRef = CI.ciSourceModule contrInfo, nmrNames = []} -- Skip finding nmrNames, as they won't be shown.
 
       let contractName = CI.getContractName contrInfo
       let wasmReceiveName = Wasm.ReceiveName [i|#{contractName}.#{receiveName}|]
-      unless (CI.hasReceiveMethod receiveName contrInfo) $ logFatal [[i|The contract does not have a receive function named '#{receiveName}'.|]]
+
+      receiveName' <- checkAndGetContractReceiveName contrInfo receiveName
```
abizjak, line 1978: I'd recommend renaming `receiveName'` to something else, or just not having the return value since it will always be the same as `receiveName`.

abizjak (commented):
Some minor tweaks are necessary, but looks good overall.

src/Concordium/Client/Output.hs
```diff
@@ -634,12 +634,12 @@ showRejectReason verbose = \case
       "account does not exist"
   Types.InvalidInitMethod m name ->
     if verbose then
-      printf "invalid init method '%s' of module '%s'" (show name) (show m)
+      printf "invalid init method %s of module '%s'" (show name) (show m)
```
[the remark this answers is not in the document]
rimbi: `printf` + `show` already puts `""` around the name, so we can remove `'`.

src/Concordium/Client/Runner.hs
rimbi: done

src/Concordium/Client/Output.hs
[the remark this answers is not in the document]
abizjak: Ok, but that is a bug since we use single quotes elsewhere and we should be consistent.

This should have single quotes only, and we should not be using show to avoid double quotes.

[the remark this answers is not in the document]
abizjak: Can you fix that as well?

src/Concordium/Client/Runner.hs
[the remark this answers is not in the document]
abizjak: Now.

src/Concordium/Client/Runner.hs
abizjak: Then please document this, and still the renaming makes sense, since `'` conveys no meaning.

What is happening here is quite subtle, since the fact that the fallback receive name is returned means that that is the name that will be used when looking up parameters/schemas. So this is quite important to get right.

src/Concordium/Client/Output.hs
```diff
@@ -634,12 +634,12 @@ showRejectReason verbose = \case
...
     else
       "invalid init method"
   Types.InvalidReceiveMethod m name ->
     if verbose then
-      printf "invalid receive method '%s' of module '%s'" (show name) (show m)
+      printf "invalid receive method %s of module '%s'" (show name) (show m)
```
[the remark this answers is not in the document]
rimbi: done

src/Concordium/Client/Runner.hs
rimbi: As far as I can see it is the convention used when a new value is assigned to a variable. Here you can see another example of it with `invoker'`. The documentation of the `checkAndGetContractReceiveName` already mentions this possibility. Do we still need to document it again here?

rimbi: The only reason to use `'` is that when I try to shadow the variable name `receiveName` the compiler gives a warning. Shadowing is something quite common in Rust, even in our sources. Apparently, one of us misses a point here. Anyway, I'll rename it.

abizjak: The reason shadowing is really bad in Haskell is that a lot of the bindings are implicitly recursive. For example
```
f x = 
   let x = someFunction x
   in someOtherFunction x
```
does not have the meaning it would have in Rust, or basically any other language.  Similarly with a `where` clause.

So that is why shadowing is a warning that is enabled by default. It is very error prone to shadow variable names in Haskell, especially when one works in a mix of settings.

One "easy" solution to this is to add `'` to variable names. But the problem with that is that it is easy to write, but the `'` has no meaning to the reader and so is hard to read, and the `'` is also easy to miss. When the code is very generic that can be fine, since a more meaningful name maybe does not exist, but in this case the code is very specific, so the variable can be named in a way that makes it easier to understand the code later on. And we should strive to have the code easy to read, and as hard to misinterpret as possible.

Some of our codebase unfortunately has `'` in the names where a better name could exist (and I am guilty of using this many times), but that is not a reason to perpetuate that bad practice.

src/Concordium/Client/Runner.hs
```diff
@@ -1859,17 +1859,19 @@ checkAndGetContractReceiveName contrInfo receiveName = do
     then return receiveName
     else
       if not $ CI.hasFallbackReceiveSupport contrInfo
-        then do
-          logWarn [[i|The contract does not have an entrypoint named '#{receiveName}' and no fallback entrypoint.|]]
-          confirmed <- askConfirmation $ Just "Do you still want to continue? "
-          unless confirmed $ logFatal ["aborting..."]
-          return receiveName
+        then confirm_abort_or_continue
```
abizjak, line 1862: The convention in Haskell, at least in our project, is to use lower camel case for variable names.

abizjak: And ideally the message in this case, and in the case below, would be slightly different.

Here the reason is that they are calling a V0 contract, below it is that they are calling a V1 contract which opts to not have a fallback.

rimbi: what about this one?

src/Concordium/Client/Commands.hs
```diff
@@ -863,7 +870,7 @@ moduleDeployCmd =
       (ModuleDeploy <$>
         strArgument (metavar "FILE" <> help "Path to the smart contract module.") <*>
         optional (strOption (long "name" <> metavar "NAME" <> help "Name for the module.")) <*>
-        optional (option (eitherReader wasmVersionFromStringInform) (long "wasm-version" <> metavar "WASM-VERSION" <> help "Optional Wasm version of the module (should only be used for modules built with cargo-concordium version < 2).")) <*>
+        optional contractVersionOption <*>
```
abizjak, line 873: This change is already on `main`. There's something off with this PR.

src/Concordium/Client/Runner.hs
```diff
@@ -1851,6 +1851,31 @@ data ModuleDeployTransactionCfg =
...
+          then do
+            logInfo [[i|The contract does not have an entrypoint named '#{receiveName}'. The fallback entrypoint will be used instead.|]]
+            return fallbackReceiveName
+          else confirmAbortOrContinue True
+  where 
+    confirmAbortOrContinue hasFallbackReceiveSupport = do
+      let fallback_msg :: String = if hasFallbackReceiveSupport
+                then " nor a fallback entrypoint"
+                else ""
+      logWarn [[i|The contract has neither an entrypoint named '#{receiveName}'#{fallback_msg}.|]]
```
abizjak, line 1874: This is not exactly the right message. If `hasFallbackReceiveSupport = False` then the message will be
```
The contract has neither an entrypoint named `foo`.
```
which is not grammatically correct.

src/Concordium/Client/Utils.hs
```diff
@@ -131,16 +131,16 @@ credentialIndexFromStringInform s =
     Nothing -> Left "Credential index must be an integer between 0 and 255 (inclusive)."
 
 -- |Try to parse a WasmVersion.
-wasmVersionFromStringInform :: String -> Either String Wasm.WasmVersion
-wasmVersionFromStringInform s =
+contractVersionFromStringInform :: String -> Either String Wasm.WasmVersion
+contractVersionFromStringInform s =
```
abizjak, line 135: This is also already on main. Something is wrong with the diff/PR.

src/Concordium/Client/Runner.hs
```diff
@@ -2121,6 +2147,7 @@ getContractUpdateTransactionCfg backend baseCfg txOpts indexOrName subindex rece
   txCfg <- getRequiredEnergyTransactionCfg baseCfg txOpts
   namedContrAddr <- getNamedContractAddress (bcContractNameMap baseCfg) indexOrName subindex
   contrInfo <- withClient backend . withBestBlockHash Nothing $ getContractInfo namedContrAddr
+  _ <- checkAndGetContractReceiveName contrInfo receiveName
   let namedModRef = NamedModuleRef {nmrRef = CI.ciSourceModule contrInfo, nmrNames = []}
   let contrName = CI.getContractName contrInfo
   schema <- withClient backend . withBestBlockHash Nothing $ getSchemaFromFileOrModule schemaFile namedModRef
```
abizjak, line 2153: (I cannot comment on the line below)

The line below, shouldn't we use `updatedReceiveName` there as well, for the same reasons this is used in `invokeContract`?

abizjak (changes requested):
There seems to be something wrong with this PR. The diff shows changes that are already in `main`.

In addition to that, there is an issue with the message, and an issue in contract update where I believe an incorrect parameter is being looked up in case of fallback.

src/Concordium/Client/Runner.hs
```diff
@@ -1851,6 +1851,31 @@ data ModuleDeployTransactionCfg =
 moduleDeployTransactionPayload :: ModuleDeployTransactionCfg -> Types.Payload
 moduleDeployTransactionPayload ModuleDeployTransactionCfg {..} = Types.DeployModule mdtcModule
 
+-- |Checks if the he given receive name is valid and if so, returns it back
```
Bargsteen, line 1854:
```suggestion
-- |Checks if the given receive name is valid and if so, returns it back
```

src/Concordium/Client/Types/Contract/Info.hs
```diff
@@ -131,6 +132,13 @@ getContractName = \case
   ContractInfoV0{..} -> ciName
   ContractInfoV1{..} -> ciName
 
+-- |Returns True if the contract has fallback entrypoint support,
+-- False otherwise
```
Bargsteen, line 136:
```suggestion
-- False otherwise.
```

src/Concordium/Client/Runner.hs
rimbi: In the `invokeContract` the client is supposed to do the fallback receiveName resolution which is supposed to be done by the node here. Isn't it?

Bargsteen: After discussing it with Cem, I made the changes you requested, Ales.

Bargsteen:
Since this is user-facing, we should add an entry in the changelog.

abizjak (approved):

PR #116, opened 2022-03-14
Adds stack.yaml file for macOS platform

rimbi:
## Purpose

Enable builds on macOS, especially with M1 chips.

## Changes

Basically overrides the compiler version and sets it to 8.10.5 which is the minimum version that exists for M1.  This is a copy of the `stack.linux-static.yaml` file with minor differences.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/stack.macos.yaml
+++ b/stack.macos.yaml
@@ -1,0 +1,34 @@
+# NB: When updating the resolver to one that uses ghc 8.10.5 or higher also
+# remove the compiler override below.
+resolver: lts-17.4
+
+# At present (Mar 14 2022) we are using a stack resolver with ghc 8.10.4 as the
+# default compiler. However we are unable to use this version of GHC because 
+# it does not exist in macOS with M1 chip. For this reason we override
+# the compiler version here. Once we update the stack resolver this should not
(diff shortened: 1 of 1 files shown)


stack.macos.yaml
```diff
@@ -0,0 +1,34 @@
+# NB: When updating the resolver to one that uses ghc 8.10.5 or higher also
+# remove the compiler override below.
+resolver: lts-17.4
+
+# At present (Mar 14 2022) we are using a stack resolver with ghc 8.10.4 as the
+# default compiler. However we are unable to use this version of GHC because 
+# it does not exist in macOS with M1 chip. For this reason we override
+# the compiler version here. Once we update the stack resolver this should not
+# be necessary anymore.
+compiler: ghc-8.10.5
```
abizjak, line 10: I do not like this addition for two reasons.
1. The same can be achieved by using a flag to `stack` when building/testing, e.g., `stack --compiler ghc-8.10.5 build`
2. It is a burden for maintenance, only relevant to 1 person at present, and is only relevant in the short term since we have to update our toolchain soon.

Hence I don't think this PR should be merged.

PR #163, opened 2022-08-05
Bump Rust version to 1.62 in build scripts.

abizjak:
## Purpose

Bump the base version in client to the latest main.

## Changes

No major changes other than version of the toolchains in build scripts.

The only code change is the removal of NonExistingRewardAccount variant which has been removed in base some time ago.

On the way, an obsolete Dockerfile has been removed. The file was clearly not used in a very long time (it uses stack resolver 16.20).

## Checklist

      hard-to-understand areas.


src/Concordium/Client/Output.hs
```diff
@@ -719,11 +719,6 @@ showRejectReason verbose = \case
   Types.RejectedReceive{..} ->
     let (contractName, funcName) = Wasm.contractAndFunctionName receiveName
     in [i|'#{funcName}' in '#{contractName}' at #{showCompactPrettyJSON contractAddress} failed with code #{rejectReason}|]
-  Types.NonExistentRewardAccount a ->
```
EmilLa1, line 722: Just to be sure - this is obsolete (and always has been on mainnet) ?

EmilLa1 (approved):
Is the Dockerfile not used anymore? I can see it used an old GHC version compared to what we use now.

src/Concordium/Client/Output.hs
abizjak: Correct. This was used a long time ago when a baker, when registering, would designate an account to be rewarded on.

This was removed at least for testnet 4 (which was the testnet around 6 months before mainnet).

abizjak:
> Is the Dockerfile not used anymore? I can see it used an old GHC version compared to what we use now.

Yeah, it was clearly not used anymore since the build just does not work, and has not for a long time. There is no reason to build concordium-client in a docker image.

PR #165, opened 2022-08-17
Fix schema parsing

### LINKED ISSUE #164: Incorrect schema parsing in concordium-client
**Bug Description**

Concordium-client does not read schemas correctly in certain cases.

In particular if the schema type contains ULEB or ILEB or ByteArray values then it parses them incorrectly.

The lenghts are parsed in big endian, not in little endian. The result of that is that the client fails to parse the supplied JSON parameters when, e.g., updating a contract.

The example schema and JSON to reproduce are part of PR #165 

The issue was first reported by @ulapcyber

abizjak:
## Purpose

Fixes #164 

## Changes

Fix parsing of schemas.

## Checklist

      hard-to-understand areas.


test/SimpleClientTests/SchemaParsingSpec.hs
```diff
@@ -0,0 +1,29 @@
...
+import qualified Data.ByteString as BS
+import qualified Data.Aeson as AE
+
+import Concordium.Client.Types.Contract.Parameter
+import Concordium.Client.Types.Contract.Schema
+
+import Test.Hspec
+import Test.HUnit
+
+-- Test that parsing a parameter that contains a set, a 32-byte array and an
```
Bargsteen, line 15: Shouldn't it be uleb128/ileb128 instead of `set`?

Bargsteen (commented):
Looks good, I just have a minor comment about the test.

test/SimpleClientTests/SchemaParsingSpec.hs
abizjak: As we discussed in the office, it is a set.

I have however added a ULEB128 to the test now as well.

Bargsteen (approved):

PR #166, opened 2022-08-19
Support V2 schemas (schemas for errors)

Bargsteen:
## Purpose

Add support for v2 schemas, which include schemas for contract errors.
Part of solving: https://github.com/Concordium/concordium-rust-smart-contracts/issues/155

## Changes

- Add the new schema (`FunctionSchemaV2`)
- Rename a number of types, by appending `V1`
- Create a number of types that are comprised of the new schemas at some level of nesting.
- Slightly simplify how module inspect info is created.

## Checklist

      hard-to-understand areas.


src/Concordium/Client/Types/Contract/Info.hs
```diff
@@ -281,10 +313,32 @@ constructModuleInspectInfo namedModRef wasmVersion moduleSchema exportedFuncName
...
+                                                                      , csv2ReceiveSigs = Map.empty
+                                                                      }) : xs
+              CS.ReceiveFuncName _ _ -> xs
+            cSigsWithoutReceives = Map.fromList . foldr mkContrSchemaTuples [] $ funcNames
+
+            insertReceiveNames :: [CS.FuncName] -> Map.Map Text ContractSigsV2 -> Map.Map Text ContractSigsV2
+            insertReceiveNames [] sigMap = sigMap
+            insertReceiveNames (CS.InitFuncName _:remaining) sigMap = insertReceiveNames remaining sigMap
+            insertReceiveNames (CS.ReceiveFuncName cname fname:remaining) sigMap = case Map.lookup cname sigMap of
+              Nothing -> insertReceiveNames remaining sigMap -- This should never happen, as we validate modules before they are put on chain.
```
abizjak, line 328: I am not sure this is correct. I don't think we do any validation on the chain for schemas, so in principle this can happen.

abizjak (approved):
Looks good.

It's a bit unfortunate we have so much code to deal with schemas, but it is what it is.

abizjak:
It would be good if you introduced a small unit test akin to the one I introduced last week.

That is, just some V2 schema with reasonable coverage (i.e., has an init, receive method, with errors.) generated with cargo-concordium and check that it parses

PR #171, opened 2022-09-26
Support importing the browser extension wallet import format.

### LINKED ISSUE #170: Support key import from the browser extension wallet
**Task description**

The browser extension wallet supports export of a single key for an account. This should be importable by concordium-client

The format of the export is
```
{
	"type": "concordium-browser-wallet-account",
	"v": 0,
	"environment": "testnet",
	"value": {
		"accountKeys": {
			"keys": {
				"0": {
					"keys": {
						"0": {
							"signKey": "...",
							"verifyKey": "..."
						}
					},
					"threshold": 1
				}
			},
			"threshold": 1
		},
		"credentials": {
			"0": "...credId..."
		},
		"address": "4SaCp3jwzP3epnV6pp4cu5Eru1nw2tWVADjjU1rsRJMbTkoHg1"
	}
}
```

The client should determine the import type from the tag, and import it. The key for encrypted transfers is not available.

abizjak:
## Purpose

Closes #170 

## Changes

Revise the mobile wallet import function to handle the new format. The new format only has one account, and does not supply the account encryption key. If such an account is imported then it will not be usable for actions that require it, such as decrypting the balance. If a user attempts this they will be met with

- [x] Depends on https://github.com/Concordium/concordium-base/pull/224

```
Error: Missing account encryption secret key for account: 3FS3Bfmds6Qj5dXHEGnzJcN8yNzfr5FqwF1qobTwYso2XxmGz2 ('account_name').
```

## Checklist

      hard-to-understand areas.

### DIFF
--- a/src/Concordium/Client/Export.hs
+++ b/src/Concordium/Client/Export.hs
@@ -75,23 +75,53 @@ data WalletExport =
   }
   deriving (Show)
 
-instance AE.FromJSON WalletExport where
-  parseJSON = AE.withObject "Wallet Export" $ \v -> do
+-- |Parse export from the wallet. The data that is exported depends a bit on
+-- which wallet it is, so the parser requires some context which is why this is
+-- a separate function, and not a @FromJSON@ instance.
+parseWalletExport ::
+  -- |The name of the account to import. The old mobile wallet export does not
+  -- require this, but for the new mobile wallet export it is required, and parsing
+  -- will fail if this is not provided.
+  Maybe Text ->
+  -- |The JSON value to be parsed.
+  AE.Value ->
+  AE.Parser WalletExport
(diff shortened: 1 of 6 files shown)


EmilLa1 (approved):
Lgtm.

src/Concordium/Client/Export.hs
```diff
@@ -141,12 +171,21 @@ decodeMobileFormattedAccountExport
   :: BS.ByteString -- ^ JSON with encrypted accounts and identities,
```
eb-concordium, line 171: The JSON is not necessarily an encryption anymore

eb-concordium (approved):
Looks good. I only have one minor comment.

src/Concordium/Client/Export.hs
abizjak: Updated documentation.

PR #179, opened 2022-10-28
Forward cookies

vikt0r0:
## Purpose

Part of the changes to allow for insertion of load-balancer between wallet-proxy endpoints. To facilitate this, user agents need to be able to include cookies in requests, and the wallet-proxy should forward these accordingly. Similarly, set-cookie headers should be retained in responses forwarded to client.

## Changes

Since the wallet-proxy relies on the the GRPC client to do so, the client was extended with functionality for passing onto it headers to be included with the request and to return headers that were included in responses.

## Checklist

      hard-to-understand areas.


src/Concordium/Client/GRPC.hs
```diff
@@ -114,6 +116,11 @@ liftClientIO comp = ClientMonad {_runClientMonad = ReaderT (\_ -> do
 runClient :: EnvData -> ClientMonad m a -> m (Either ClientError a)
 runClient config comp = runExceptT $ runReaderT (_runClientMonad comp) config
 
+runClientWithExtraHeaders :: GRPCHeaderList -> EnvData -> ClientMonad m a -> m (Either ClientError a)
+runClientWithExtraHeaders hds cfg comp = runExceptT $ runReaderT (_runClientMonad comp) cfgWithHeaders
+  where headerList = Prelude.map (\(k,v) -> (CI.original k, v)) hds
+        cfgWithHeaders = cfg { config = (config cfg) { _grpcClientConfigHeaders = _grpcClientConfigHeaders (config cfg) ++ headerList } }
```
abizjak, line 122: Please add documentation.

src/Concordium/Client/Runner/Helper.hs
```diff
@@ -69,3 +78,6 @@ value s =
   case eitherDecodeStrict' s of
     Right v -> v
     Left err -> error ("Error in gRPC output decoding as a json: " ++ err)
+
+value' :: forall a b. (Field.HasField a "value" b) => GRPCResponse a -> GRPCResponse b
+value' = fmap (^. CF.value)
```
abizjak, line 83: It is unclear what this is for. The name is certainly not very meaningful.

It would be better to have something like
```haskell
-- |Extract a value from the response. This assumes that the response from a GRPC call has a @value!@ field.
getValue :: forall a b. (Field.HasField a "value" b) => SimpleGetter (GRPCResponse a) (GRPCResponse b)
getValue = to (fmap (^. CF.value))
```

src/Concordium/Client/Types/GRPC.hs
```diff
@@ -0,0 +1,20 @@
...
+data GRPCResponse a = GRPCResponse
+  { grpcHeaders :: CIHeaderList,
+    grpcResponse :: a
+  } deriving Show
+
+instance Functor GRPCResponse where
+  fmap f (GRPCResponse hds v) = GRPCResponse hds $ f v
+
+type GRPCResult a = Either String (GRPCResponse a)
+type GRPCHeaderList = CIHeaderList
```
abizjak, line 20: Add this go `GRPC` directly and remove this module. We already have many types there so let's keep that.

src/Concordium/Client/Runner/Helper.hs
```diff
@@ -37,30 +41,35 @@ outputGRPC ret =
...
     Left e -> Left $ "Unable to send consensus query: " ++ show e
 
-
 processJSON :: (Field.HasField a "value" Text) => a -> Value
 processJSON val = do
   let r = val ^. CF.value
   value . encodeUtf8 $ r
 
+processJSON' :: (Field.HasField a "value" Text) => GRPCResponse a -> GRPCResponse Value
+processJSON' = fmap processJSON
```
abizjak, line 62: Document.

src/Concordium/Client/Runner.hs
```diff
@@ -161,6 +163,9 @@ getFromJson :: (MonadIO m, FromJSON a) => Either String Value -> m a
 getFromJson = getFromJsonAndHandleError onError
   where onError val err = logFatal ["cannot convert '" ++ show val ++ "': " ++ err]
 
+getFromJson' :: (MonadIO m, FromJSON a) => GRPCResult Value -> m a
+getFromJson' = getFromJson . fmap grpcResponse
+
```
abizjak, line 168: Document. The name is not meaningful.

src/Concordium/Client/Types/GRPC.hs
```diff
@@ -0,0 +1,20 @@
...
+
+import Network.GRPC.Client
+
+data GRPCResponse a = GRPCResponse
+  { grpcHeaders :: CIHeaderList,
+    grpcResponse :: a
+  } deriving Show
+
+instance Functor GRPCResponse where
+  fmap f (GRPCResponse hds v) = GRPCResponse hds $ f v
```
abizjak, line 17: This can be derived, and it should be 
```
deriving (Functor)
```

src/Concordium/Client/Types/GRPC.hs
```diff
@@ -0,0 +1,20 @@
...
+  , GRPCResponse(..)
+  , GRPCHeaderList
+  ) where
+
+import Network.GRPC.Client
+
+data GRPCResponse a = GRPCResponse
+  { grpcHeaders :: CIHeaderList,
+    grpcResponse :: a
+  } deriving Show
```
abizjak, line 14: Documentation.

src/Concordium/Client/Types/GRPC.hs
abizjak, line 20: Documentation.

abizjak (changes requested):

src/Concordium/Client/Types/GRPC.hs
vikt0r0: I did this initially but this introduces a cyclic dependency between GRPC and Helper. Is there some way to avoid that?

abizjak: I see.

We could add it to Helper then, or move the helper definitions that need this to GRPC. We only need those internally anyhow.

vikt0r0:
I addressed review comments.

abizjak:
Please address the remaining comments, then merge.

PR #184, opened 2022-11-05
Support for Schema V3 in "module inspect", "contract show" and "transaction status".

vikt0r0:
## Purpose

Add support for V3 schemas. This entails implementing (de)serialization of event schemas and (de)serialization of events using schemas as well as outputting it in nice, human friendly JSON.

## Changes

- Add "tagged enum" type to SchemaType and implement (de)serializing and JSON printing.
- Add support for (de)serializing events w.r.t. the updated SchemaType.
- Add contract/module V3 schema datatypes.
- Extend `contract show` and `module inspect` to allow for printing V3 schemas.
- Extend `transaction status` to allow for printing event information.
- Print events in `tailTransaction`.
- Add tests for deserializing/serializing.

## Checklist

      hard-to-understand areas.
- [X] I have updated the CHANGELOG.


vikt0r0:
Doris kindly provided a V3 contract to test this with:

`concordium-client --grpc-ip <stagenet> --grpc-port 10000 transaction status 4e1c2871d66b79d67fdd8fa6ea557b31866970c5f424ad965540fc748092c8df`

`concordium-client --grpc-ip <stagenet> --grpc-port 10000 module inspect cdb2747c27d1921240e8ce4c9d638cc7e887e4df40d2a253e9bc5454c51dcb73`

`concordium-client --grpc-ip <stagenet> --grpc-port 10000 contract show 1025`

src/Concordium/Client/Types/Contract/Schema.hs
```diff
@@ -449,6 +509,9 @@ instance AE.ToJSON SchemaType where
     ILeb128 _ -> AE.String "<String with signed integer>"
     ByteList _ -> AE.String "<String with lowercase hex>"
     ByteArray _ -> AE.String "<String with lowercase hex>"
+    TaggedEnum variants -> AE.object ["TaggedEnum" .=
+      (toJsonArray . map (\(k, v) -> AE.object [AE.fromText k .= v]) $
+      (\(k,v) -> (pack $ show k, v)) <$> Map.toList variants)]
```
abizjak, line 514: Can you please write this in a more readable way. Parsing this is terribly complicated now. Name the intermediate values at least.
What is the intended display here even? I would have thought the tags should be ignored, since they are not relevant for what the user needs to write.

abizjak (changes requested):

src/Concordium/Client/Types/Contract/Schema.hs
vikt0r0: The intention is to display the TaggedEnum as JSON. I suppose whether the tag should be displayed depends on the semantics and how the client is meant to be used. 

If the string is a key I guess it does not have to, if not, I think it better be present. Tags are also displayed in the output of concordium-cargo to the my best knowledge.

src/Concordium/Client/Output.hs
```diff
@@ -699,6 +735,19 @@ showEvent verbose = \case
     showDelegationTarget Types.DelegatePassive = "Passive delegation"
     showDelegationTarget (Types.DelegateToBaker bid) = "Baker ID " ++ show bid
 
+    showLoggedEvents :: Maybe SchemaType -> [Wasm.ContractEvent] -> String
+    showLoggedEvents st evs = "\n" <> case evs of
+      [] -> "No contract events were logged in this transaction."
```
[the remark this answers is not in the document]
vikt0r0: How should they differ.

To my knowledge the output of `show transaction` already states the event type. `showLoggedEvents` is currently only called from `showEvent` in case of `ContractInitialized`, `Interruped` or `Update`.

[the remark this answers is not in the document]
abizjak: For the `ContractInitialized` this message is fine.

For when this is applied to the `Interrupt` then it is not correct. The logs in that event are only for the section of the code before the interrupt. It is a bit hard to state this very precisely since the exact condition is complex, but something like `"No contract events were emitted." is probably fine.

I would not mention "transaction" for Interrupted and Updated events.

src/Concordium/Client/Types/Contract/Schema.hs
abizjak: In what context are the tags printed?

As the comment above states, the output here should be roughly what the user is expected to write to supply the value.

As far as I see in the cargo-concordium code the user is meant to supply the same format as in the `Enum` case, so no tags.

Can you please make sure that this agrees?

src/Concordium/Client/Types/Contract/Parameter.hs
```diff
@@ -244,6 +256,23 @@ putJSONUsingSchema typ json = case (typ, json) of
         pure $ putLen <> putJSONFields'
     _ -> Left [i|#{obj} had too many fields. It should contain a single variant of the following enum:\n#{showPrettyJSON enum}.|]
 
+  (tEnum@(TaggedEnum variants), AE.Object obj) -> case KM.toList obj of
```
Bargsteen, line 259: I want to have a look at this again when it has been aligned with cargo-concordium 😊

src/Concordium/Client/Types/Contract/Schema.hs
```diff
@@ -195,6 +232,17 @@ data ContractSchemaV2
 
 instance AE.ToJSON ContractSchemaV2
 
+-- |Parallel to schema::ContractV3 defined in concordium-contracts-common (Rust) version > 2.
```
Bargsteen, line 235:
```suggestion
-- |Parallel to schema::ContractV3 defined in concordium-contracts-common (Rust) version >= 5.
```

src/Concordium/Client/Types/Contract/Schema.hs
```diff
@@ -195,6 +232,17 @@ data ContractSchemaV2
...
+data ContractSchemaV3
+  = ContractSchemaV3 -- ^ Describes the schemas of a V1 smart contract.
+  { cs3InitSig :: Maybe FunctionSchemaV2 -- ^ Schema for the init function.
+  , cs3ReceiveSigs :: Map Text FunctionSchemaV2 -- ^ Schemas for the receive functions.
+  , cs3EventSchema :: Maybe SchemaType -- ^ Schemas for the events
```
Bargsteen, line 240:
```suggestion
  , cs3EventSchema :: Maybe SchemaType -- ^ Schemas for the events.
```

src/Concordium/Client/Types/Contract/Schema.hs
```diff
@@ -282,6 +338,9 @@ data FunctionSchemaV2
                  }
   deriving (Eq, Show, Generic)
 
+-- |V3 Schema for events in a V1 smart contract
```
Bargsteen, line 341:
```suggestion
-- |V3 Schema for events in a V1 smart contract.
```

src/Concordium/Client/Types/Contract/Schema.hs
Bargsteen: I haven't tested cargo-concordium with the event schemas, but it should be sufficient to provide the variant name and then use that to find the index in the schema.

src/Concordium/Client/Types/Contract/Schema.hs
```diff
@@ -217,6 +265,14 @@ instance S.Serialize ContractSchemaV2 where
     pure ContractSchemaV2{..}
   put ContractSchemaV2 {..} = S.put cs2InitSig <> putMapOfWithSizeLen Four putText S.put cs2ReceiveSigs
 
+instance S.Serialize ContractSchemaV3 where
```
Bargsteen, line 268: Please add some tests cases to 
`/test/SimpleClientTests/SchemaParsingSpec.hs` for this.

src/Concordium/Client/Types/Contract/Parameter.hs
```diff
@@ -244,6 +256,23 @@ putJSONUsingSchema typ json = case (typ, json) of
...
+    [] -> Left [i|The object provided was empty, but it should have contained a variant of the following tagged enum:\n#{showPrettyJSON tEnum}.|]
+    [(tag, val@(AE.Array vec))] -> case AE.toString tag of
```
Bargsteen, line 261: Please add some test cases for this in `test/SimpleclientTests/ContractSpec.hs`

Bargsteen (changes requested):
Nice work! It's nearly there, just missing a few details ;)

src/Concordium/Client/Types/Contract/Schema.hs
vikt0r0: I'll have a look at cargo-concordium and make it is consistent with how it's done there.

src/Concordium/Client/Output.hs
```diff
@@ -465,6 +484,23 @@ printModuleInspectInfo CI.ModuleInspectInfo{..} = do
...
+    showContractSigsV3 = go . sortOn fst . Map.toList
+      where go [] = []
+            go ((cname, CI.ContractSigsV3{..}):remaining) = [showContractFuncV2 cname csv3InitSig]
+                                                          ++ showReceives (sortOn fst . Map.toList $ csv3ReceiveSigs)
+                                                          ++ showEvents cs3EventSchema
+                                                          ++ go remaining
+
+            showReceives :: [(Text, Maybe CS.FunctionSchemaV2)] -> [String]
+            showReceives [] = []
+            showReceives ((fname, mSchema):remaining) = indentBy 4 (showContractFuncV2 fname mSchema) : showReceives remaining
```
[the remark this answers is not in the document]
vikt0r0: Good point! I actually wanted to do something like that, but since it was not already done, I adhered to following the style of the code. Is such refactoring encouraged, however?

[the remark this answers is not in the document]
Bargsteen: The v2 code was written in quite a rush (by me), so that is why it wasn't done. Or perhaps I simply overlooked it ;) 
Well, that depends. In this simple and isolated case, I think it's fine, but refactoring should provide a decent amount of benefits before it is worth it. Because it is not just the refactoring itself that takes time. There is also the time needed for reviewing the changes, retesting the changes, and time needed for other developers to learn the new API or semantics.
So if in doubt, ask ;)

src/Concordium/Client/Output.hs
```diff
@@ -699,6 +735,19 @@ showEvent verbose = \case
     showDelegationTarget Types.DelegatePassive = "Passive delegation"
     showDelegationTarget (Types.DelegateToBaker bid) = "Baker ID " ++ show bid
 
+    showLoggedEvents :: Maybe SchemaType -> [Wasm.ContractEvent] -> String
+    showLoggedEvents st evs = "\n" <> case evs of
+      [] -> "No contract events were logged in this transaction."
+      _ -> case st of 
+        Nothing -> [i|#{length evs} contract events were logged. No V3 event schema was found in contract nor provided.|]
+        Just st' -> do
+          let (_, vals) =
```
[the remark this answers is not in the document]
vikt0r0: Should we display HEX and JSON in case a schema's present?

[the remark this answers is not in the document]
Bargsteen: In my mind, this is what we should do:

1) No schema => Display hex
2) Schema with at least 1 failed decoding => Display hex + warning about incorrect schema
3) Schema with succesful decoding => Display JSON

One could argue that in second case, we should display the successfully decoded events as JSON and the rest as hex. That could kind of make sense, fx if you are using an old schema which only has 4 event types, but you updated the contract so now it now has 5 event types. But I think the risk of displaying incorrect JSON is too large, so I would rather just display the hex in this case.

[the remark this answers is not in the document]
abizjak: Case 2 is a bit tricky.

A reasonable thing to do **might** be to only include the schema for some events, e.g., CIS2 events, but not for any other ones you are emitting.

It is ultimately the contract writer who embeds the schema, so I think we should try to just decode (the entire log item!) and output JSON if we can. That is also a much more local behaviour that is easier to implement that waiting for all the events to be decoded before we can output any of them.

src/Concordium/Client/Output.hs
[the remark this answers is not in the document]
vikt0r0: Makes sense :-) I agree refactoring makes sense here, and I'll add it here, and in the future I'll ask if I'm in doubt 👍

src/Concordium/Client/Runner.hs
```diff
@@ -570,13 +571,38 @@ processTransactionCmd action baseCfgDir verbose backend =
...
+      let cAddr = extractFromTsr (\case Types.ContractInitialized {..} -> Just ecAddress
+                                        Types.Updated {..} -> Just euAddress
+                                        Types.Interrupted {..} -> Just iAddress
+                                        _ -> Nothing) $ Just status
+      schema <- case cAddr of
+        Just (Right ca) -> do
+          let caT = (Text.pack . show . Types.contractIndex) ca
+          baseCfg <- getBaseConfig baseCfgDir verbose
+          namedContrAddr <- getNamedContractAddress (bcContractNameMap baseCfg) caT Nothing
+          withClient backend . withBestBlockHash Nothing $ \bb -> do
```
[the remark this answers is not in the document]
vikt0r0: Do you have any pointers as to how to determine the block used? TransactionBlockStatus has several cases, but I am unsure of the semantics.

src/Concordium/Client/Output.hs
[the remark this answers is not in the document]
Bargsteen: I am not sure what you mean about waiting for all events. I meant all the events given in the `[Wasm.ContractEvent]`.

vikt0r0:
Hopefully this looks alright now. I included some output here for convenience.

[contract_show.txt](https://github.com/Concordium/concordium-client/files/996
[transaction_status.txt](https://github.com/Concordium/concordium-client/files/9968504/transaction_status.txt)
8338/contract_show.txt)
[module_inspect.txt](https://github.com/Concordium/concordium-client/files/9968339/module_inspect.txt)

src/Concordium/Client/Output.hs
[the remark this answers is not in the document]
Bargsteen: Ah ok. Yea, I just meant a smaller scope ;)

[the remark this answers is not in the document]
Bargsteen: Ah okay. Yea I just meant a smaller scope ;)

src/Concordium/Client/Types/Contract/Parameter.hs
```diff
@@ -244,6 +255,18 @@ putJSONUsingSchema typ json = case (typ, json) of
         pure $ putLen <> putJSONFields'
     _ -> Left [i|#{obj} had too many fields. It should contain a single variant of the following enum:\n#{showPrettyJSON enum}.|]
 
+  (tEnum@(TaggedEnum variants), AE.Object obj) -> case KM.toList obj of
+    [] -> Left [i|The object provided was empty, but it should have contained a variant of the following tagged enum:\n#{showPrettyJSON tEnum}.|]
+    [(name, fields)] -> case lookupItemAndIndex name $ map (first AE.fromText . snd) variantList of
+      Nothing -> Left [i|Tagged enum variant '#{name}' does not exist in:\n#{showPrettyJSON tEnum}|]
+      Just (fieldTypes, idx) -> do
+        let tag = S.putWord8 $ fst $ variantList !! fromIntegral idx
+        putJSONFields' <- putJSONFields fieldTypes fields `addTraceInfoOf` [i|In enum variant '#{name}'.|]
```
Bargsteen, line 264:
```suggestion
        putJSONFields' <- putJSONFields fieldTypes fields `addTraceInfoOf` [i|In tagged enum variant '#{name}'.|]
```

src/Concordium/Client/Types/Contract/Schema.hs
```diff
@@ -716,8 +790,21 @@ putLenWithSizeLen sl len = case sl of
...
 -- * Map *
 
-getMapOfWithSizeLen :: Ord k => SizeLength -> S.Get k -> S.Get v -> S.Get (Map k v)
-getMapOfWithSizeLen sl gt gv = S.label "Map" $ Map.fromList <$> getListOfWithSizeLen sl (S.getTwoOf gt gv)
+allUnique :: (Ord a) => [a] -> Bool
+allUnique xs = length (nub xs) /= length xs
+
+-- |Get a map with a specied size length. Fails if each element in the list of
+-- deserialized map tuples mapped over f is not unique.
+getMapOfWithSizeLenAndPred :: (Ord k) => ([(k,v)] -> Bool) -> SizeLength -> S.Get k -> S.Get v -> S.Get (Map k v)
```
Bargsteen, line 798: This mentions uniqueness, but the predicate could be used for checking other properties.
I would suggest describing that it takes a predicate for checking some property and that this fx could be uniqueness.
Please also update the error message a few lines below with the same idea in mind.

src/Concordium/Client/Output.hs
```diff
@@ -699,6 +727,23 @@ showEvent verbose = \case
...
     showDelegationTarget (Types.DelegateToBaker bid) = "Baker ID " ++ show bid
 
+    showLoggedEvents :: [Wasm.ContractEvent] -> String
+    showLoggedEvents [] = "No contract events were emitted "
+    showLoggedEvents evs = [i|#{length evs} contract events were emitted|]
+      <> if isNothing stM
+         then [i| but no event schema was provided nor found in the contract module. |]
+         else [i|, of which #{length $ filter isJust $ map snd loggedEvents} were succesfully parsed using schema. |]
+      <> [i|Got:\n|]
+      <> intercalate "\n" [ [i|Event(raw): #{hex}\nEvent(JSON):\n#{indentBy 4 $ fromMaybe' json}|] | (hex, json) <- loggedEvents ]
```
Bargsteen, line 737: I don't think we need to print the hex value if we decode it successfully.
So I would suggest to just print one or the other.

Bargsteen (approved):
Looks good, just a few minor issues.

But you might still want to add a test case for the new schema in `/test/SimpleClientTests/SchemaParsingSpec.hs`

src/Concordium/Client/Output.hs
vikt0r0: @abizjak what do you think?

src/Concordium/Client/Commands.hs
```diff
@@ -678,7 +680,8 @@ transactionStatusCmd =
     "status"
     (info
       (TransactionStatus <$>
-        strArgument (metavar "TX-HASH" <> help "Hash of the transaction."))
+        strArgument (metavar "TX-HASH" <> help "Hash of the transaction.") <*>
+        optional (strOption (long "schema" <> metavar "SCHEMA" <> help "Path to a schema file, used to display the contract info.")))
```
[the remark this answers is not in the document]
abizjak: @vikt0r0 You have resolved this conversation, but you have not fixed the issue. Please fix the help text so that it is correct.

src/Concordium/Client/Output.hs
```diff
@@ -381,6 +396,10 @@ showContractFuncV2 funcName mFuncSchema = case mFuncSchema of
   Just CS.RvError{..} -> [i|- #{funcName}\n    Return value:\n#{indentBy 8 $ showPrettyJSON fs2ReturnValue}\n    Error:\n#{indentBy 8 $ showPrettyJSON fs2Error}|]
   Just CS.ParamRvError{..} -> [i|- #{funcName}\n    Parameter:\n#{indentBy 8 $ showPrettyJSON fs2Parameter}\n    Return value:\n#{indentBy 8 $ showPrettyJSON fs2ReturnValue}\n    Error:\n#{indentBy 8 $ showPrettyJSON fs2Error}|]
 
+showContractEventV3 :: Maybe SchemaType -> String
```
abizjak, line 399: Add short comment. 

Just because existing functions do not have them does not mean new ones should not either.

src/Concordium/Client/Output.hs
```diff
@@ -591,16 +619,17 @@ showOutcomeResult verbose = \case
...
-showEvent verbose = \case
+showEvent :: Verbose -> Maybe SchemaType -> Types.Event -> Maybe String
+showEvent verbose stM = \case
   Types.ModuleDeployed ref->
     verboseOrNothing $ printf "module '%s' deployed" (show ref)
   Types.ContractInitialized{..} ->
-    verboseOrNothing $ printf "initialized contract '%s' using init function '%s' from module '%s' with %s"
-                              (show ecAddress) (show ecInitName) (show ecRef) (showCcd ecAmount)
+    verboseOrNothing $ [i|initialized contract '#{ecAddress}' using init function '#{ecInitName}' from module '#{ecRef}' |]
+                    <> [i|with #{ecAmount}\n#{showLoggedEvents ecEvents}|]
```
abizjak, line 628: Please use `showCcd` and **not** the Show instance of Amount. The former renders it in a uniform way everywhere in the client. The latter is the simple show instance used for debugging.

Also generally do not do refactoring unless necessary. It is very easy to make simple mistakes like this.

src/Concordium/Client/Output.hs
```diff
@@ -699,6 +728,26 @@ showEvent verbose = \case
...
     showDelegationTarget (Types.DelegateToBaker bid) = "Baker ID " ++ show bid
 
+    showLoggedEvents :: [Wasm.ContractEvent] -> String
+    showLoggedEvents [] = "No contract events were emitted "
+    showLoggedEvents evs = [i|#{length evs} contract events were emitted|]
+      <> if isNothing stM
+         then [i| but no event schema was provided nor found in the contract module. |]
+         else [i|, of which #{length $ filter isJust $ map snd loggedEvents} were succesfully parsed using schema. |]
+      <> [i|Got:\n|]
+      <> intercalate "\n" [ if isJust json
```
abizjak, line 738: Use `unlines` instead.

src/Concordium/Client/Output.hs
```diff
@@ -699,6 +728,26 @@ showEvent verbose = \case
...
+                            then [i|Event(parsed):\n#{indentBy 4 $ fromMaybe' json}|]
+                            else [i|Event(raw): #{hex}\n|]
+                            | (hex, json) <- loggedEvents ]
+        where
+          fromMaybe' :: Maybe AE.Value -> String
+          fromMaybe' = \case Nothing -> "Unable to decode."; Just j -> showSortedPrettyJSON j
```
abizjak, line 744: Please do not introduce new syntax. We do not use semicolons in Haskell to squeeze things into a single line.

Use the prevailing code style. The simpler the better.

src/Concordium/Client/Types/Contract/Parameter.hs
```diff
@@ -44,6 +44,10 @@ import GHC.Integer (remInteger, modInteger)
 encodeParameter :: SchemaType -> AE.Value -> Either String ByteString
 encodeParameter typ params = S.runPut <$> putJSONUsingSchema typ params
 
+-- |Deserialize binary to JSON using `SchemaType` or fail with an error message.
+decodeParameter :: SchemaType -> ByteString -> Either String AE.Value
+decodeParameter typ = S.runGet $ getJSONUsingSchema typ
```
abizjak, line 49: The name of this is rather odd, since it has nothing to do with parameters. It is used for events as well.

abizjak: Also, for decoding events it is **essentiall** that the entire event is consumed while parsing. Otherwise we could easily have buggy, misleading prints.

So we have to make sure that `getJSONUsingSchema`  consumes the entire input.

src/Concordium/Client/Output.hs
```diff
@@ -699,6 +728,26 @@ showEvent verbose = \case
...
+          toHex :: Wasm.ContractEvent  -> String
+          toHex (Wasm.ContractEvent bs) = show . BSB.toLazyByteString . BSB.byteStringHex $ SE.runPut $ SE.putShortByteString bs
+          toJSON' :: Wasm.ContractEvent -> Maybe AE.Value
+          toJSON' (Wasm.ContractEvent bs) =
+            stM >>= \st' -> (preview _Right . PA.decodeParameter st' . SE.runPut . SE.putShortByteString) bs
```
abizjak, line 749: Can you honestly say that in 1month this code will be readable? 

Please just do a simple thing. A line or two more is better than weird sequences of function compositions with hard to understand intermediate results.
```
-- Attempt to decode the contract event if the schema is provided.
-- If there is no schema, or decoding fails @Nothing@ is returned.
toJSON' (Wasm.ContractEvent bs) = do
    st' <- stM
    case PA.decodeParameter (BSS.fromShort st') of
        Left _ -> Nothing
        Right x -> Just x
```

src/Concordium/Client/Output.hs
```diff
@@ -699,6 +728,26 @@ showEvent verbose = \case
...
+          loggedEvents = map (bimap toHex toJSON') (zip evs evs)
```
abizjak, line 750: This is a very complex way of writing it and I do not see why it needs to be. Why can't we just do

```haskell
eventToString e = case toJSON' e of
    Nothing -> [i|Event(raw): #{toHex e}\n|]
    Just json -> [i|Event(parsed):\n#{indentBy 4 $ json}|]
```

And then instead of the `intercalate` just to `unlines . map eventToString $ evs` or similar.

It'll also remove the strange `fromMaybe'`.

src/Concordium/Client/Output.hs
abizjak: This is also a very strange function as is. And if it remains needs to have rationale in its documentation.

src/Concordium/Client/Runner.hs
```diff
@@ -570,13 +592,16 @@ processTransactionCmd action baseCfgDir verbose backend =
...
       hash <- case parseTransactionHash h of
         Nothing -> logFatal [printf "invalid transaction hash '%s'" h]
         Just hash -> return hash
-      status <- withClient backend $ queryTransactionStatus hash
-      when verbose $ logInfo [printf "Response: %s" (show status)]
-      runPrinter $ printTransactionStatus status True
+      withClient backend . withBestBlockHash Nothing $ \bb -> do
+        status <- queryTransactionStatus hash
+        schema <- getEventSchema verbose schemaFile status $ Just bb
+        runPrinter $ printTransactionStatus status True schema
```
abizjak, line 603: The schema should not be from the best block. It should be from the block that the transaction status returns.

The transaction status is a mapping from block hashes to outcomes. We need to get the relevant schema out.

While using the "best" block will almost always be correct, it will not **always** be correct.

src/Concordium/Client/Runner.hs
```diff
@@ -544,6 +544,28 @@ checkAndGetMemo memo pv = do
       return $ Types.Memo bss
     Right m -> return m
 
+-- |Get the contract V3 event schema associated with a transaction if present.
+getEventSchema :: (MonadIO m, MonadFail m) => Bool -> Maybe FilePath -> TransactionStatusResult -> Maybe Text -> ClientMonad m (Maybe CS.SchemaType)
```
abizjak, line 548: Please document the meaning of parameters.

src/Concordium/Client/Runner.hs
```diff
@@ -1379,7 +1404,9 @@ tailTransaction hash = do
     logFatal [ "transaction failed before it got committed"
              , "most likely because it was invalid" ]
 
-  runPrinter $ printTransactionStatus committedStatus False
+  schema <- getEventSchema False Nothing committedStatus Nothing
```
abizjak, line 1407: Didn't the user set the `verbose` flag somewhere before they got to here? WHy are we setting it to `False`?

src/Concordium/Client/Types/Contract/Info.hs
```diff
@@ -472,3 +545,11 @@ data ContractSigsV2
   { csv2InitSig :: Maybe CS.FunctionSchemaV2 -- ^ Schema for the init function.
   , csv2ReceiveSigs :: Map.Map Text (Maybe CS.FunctionSchemaV2) -- ^ Schemas for the receive functions.
   }
+
+-- |Different from ContractSchemaV3 in that the receiveSigs have a Maybe FunctionSchemaV2.
```
abizjak, line 549: This comment is not correct.

src/Concordium/Client/Types/Contract/Schema.hs
```diff
@@ -123,13 +154,20 @@ lookupFunctionSchemaV2 ContractSchemaV2{..} = \case
   InitFuncName _ -> cs2InitSig
   ReceiveFuncName _ rcvName -> Map.lookup rcvName cs2ReceiveSigs
 
+-- |Lookup the 'FunctionSchemaV2' of a 'ContractSchemaV3'.
```
abizjak, line 157:
```suggestion
-- |Look up either the schema of an init or a receive function.
```

src/Concordium/Client/Types/Contract/Schema.hs
```diff
@@ -486,7 +553,15 @@ instance S.Serialize SchemaType where
       28 -> S.label "ILeb128"  $ ILeb128 <$> S.getWord32le
       29 -> S.label "ByteList" $ ByteList <$> S.get
       30 -> S.label "ByteArray"  $ ByteArray <$> S.getWord32le
+      31 -> S.label "TaggedEnum" $ TaggedEnum
+        <$> getMapOfWithSizeLenAndPred tEnumPred Four S.getWord8 (S.getTwoOf getText S.get)
       x  -> fail [i|Invalid SchemaType tag: #{x}|]
+    where
+      -- Predicate for tagged enums. Tags and variant names should be unique.
+      allUnique :: (Ord a) => [a] -> Bool
+      allUnique xs = length (nub xs) == length xs
```
abizjak, line 562: Nub is quadratic, but you required a to be comparable. So use sort sort + group instead.

src/Concordium/Client/Types/Contract/Schema.hs
```diff
@@ -716,9 +792,19 @@ putLenWithSizeLen sl len = case sl of
...
 -- * Map *
 
-getMapOfWithSizeLen :: Ord k => SizeLength -> S.Get k -> S.Get v -> S.Get (Map k v)
-getMapOfWithSizeLen sl gt gv = S.label "Map" $ Map.fromList <$> getListOfWithSizeLen sl (S.getTwoOf gt gv)
+-- |Get a map with a specified size length. Fails if the predicate is false when applied to the list of tuples in the map.
+getMapOfWithSizeLenAndPred :: (Ord k) => ([(k,v)] -> Bool) -> SizeLength -> S.Get k -> S.Get v -> S.Get (Map k v)
+getMapOfWithSizeLenAndPred p sl gt gv = do
+  ls <- getListOfWithSizeLen sl (S.getTwoOf gt gv)
+  if p ls
+  then S.label "Map" $ pure $ Map.fromList ls
```
abizjak, line 800: I am not sure what the `label` there achieves. This will never fail so I don't think the label will ever be printed.

IN general, this is also not the way to do things. You have in essence made all the `getMapOfWithSizeLen` slower. In practice this won't matter **in the client for the specific case we have**, since the amount of data is sufficiently small, but we do need to be careful when doing things like this.

test/SimpleClientTests/SchemaParsingSpec.hs
```diff
@@ -51,3 +55,52 @@ schemaV2Parsing = specify "Parse v2 schema" $ do
         Just schema -> case putJSONUsingSchema schema value of
           Left err -> assertFailure $ "Could not convert JSON according to the provided schema: " ++ err
           Right _ -> return ()
+
+
+-- | Module schema V3 tests 
+prefix :: String
```
abizjak, line 61:
```suggestion
prefix :: Filepath
```

test/SimpleClientTests/SchemaParsingSpec.hs
```diff
@@ -51,3 +55,52 @@ schemaV2Parsing = specify "Parse v2 schema" $ do
...
+prefix = "test/data/cis2_wCCD/"
+
+getSchema :: IO SchemaType
```
abizjak, line 64: Document.

abizjak (changes requested):

src/Concordium/Client/Runner.hs
vikt0r0: It was set to false and thought this was the intended behaviour. I'll change it to match the value of the command line parameter?

src/Concordium/Client/Types/Contract/Info.hs
vikt0r0: I think it's correct albeit I do not particularly like the wording of the comments for the other schemas, which this comment follows. But I wanted to be consistent. Here's ContractSchemaV3 for the record:

```
data ContractSchemaV3
  = ContractSchemaV3 -- ^ Describes the schemas of a V1 smart contract.
  { cs3InitSig :: Maybe FunctionSchemaV2 -- ^ Schema for the init function.
  , cs3ReceiveSigs :: Map Text FunctionSchemaV2 -- ^ Schemas for the receive functions.
  , cs3EventSchema :: Maybe SchemaType -- ^ Schemas for the events.
  }
  deriving (Eq, Show, Generic)
```

Should I remove the comment?

src/Concordium/Client/Types/Contract/Parameter.hs
vikt0r0: I agree. The same goes for the already existing `encodeParameter` as well, but I wanted to follow the style of the current naming. 

This seems to be an issue of more general nature, so moving forward should I a) rename oddly named functions or b) rather name the ones I add suitably, even though there may be related functions with odd names?

I'll add the end-of-input check for getJSONUsingSchema 👍🏼

src/Concordium/Client/Types/Contract/Schema.hs
vikt0r0: True, but I thought this was much more readable since the asymptotics don't matter much here (as n <= 256). I'll change it.

src/Concordium/Client/Types/Contract/Schema.hs
vikt0r0: Label is there since I assumed that the code I modified was sound by induction. Should I remove it?

What is making this so terribly slow, and what is the better way of accomplishing this?  I hope and pray that GHC is capable of inferring that

`getMapOfWithSizeLenAndPred (const True) sl gt gv` \equiv `\sl gt gv -> S.label "Map" $ Map.fromList <$> getListOfWithSizeLen sl (S.getTwoOf gt gv)`

Maybe there is something I misunderstand :-)

src/Concordium/Client/Output.hs
vikt0r0: I agree. This function as a whole is terrible, and I'll rewrite it.

src/Concordium/Client/Output.hs
Bargsteen: I'm to blame for the small refactoring, as I suggested it.

src/Concordium/Client/Types/Contract/Parameter.hs
Bargsteen: Yes, this is a legacy from the time when we only had schemas for parameters.
I think we should rename them (and the file), but perhaps that should be done in a separate PR.

src/Concordium/Client/Runner.hs
Bargsteen: IIRC there was a reason for always using False, but I don't recall what it was. Let's use what the user specified, but please check that the output seems reasonable.

src/Concordium/Client/Types/Contract/Info.hs
Bargsteen: In terms of the types, the comment is correct. But I guess we could add some explanation as to why this differs.
It is because we want to list all the receive functions, and some of them we have more data for via the schema.

src/Concordium/Client/Types/Contract/Schema.hs
abizjak: It is not that it is **terribly** slow, and as I wrote in my comment, in practice it most likely will not be observable.

But it is slower than it was.
- First a list is constructed
- Then a list is traversed, applying a function to it.
- Then the list is converted to a map.

You might say "laziness makes this OK" and it is true, **it might**.

I would not use this function to implement `getMapOfWithSizeLen`. But again, since the number of elements here will be at most a few thousand it will not be noticeable and this is not worth more of a discussion.

My general point is that abstractions have a cost and we should be mindful of the cost at every level and not make things needlessly slower.

src/Concordium/Client/Types/Contract/Info.hs
abizjak: Right.

Also, such a comment is pretty bad since it says nothing useful. It does not describe a purpose of the type which is what a comment should generally do.

And I fully realize that many comments are not like that. That is not the reason to perpetuate the badness.

Please add as Kasper suggests.

abizjak (approved):

PR #189, opened 2022-11-14
Improve disconnect detection and GOAWAY handling.

### LINKED ISSUE #188: Connection reset is not handled promptly by the client, middleware, wallet-proxy
**Bug Description**

The internal network client used by concordium-client, the wallet-proxy, and the middleware does not handle GOAWAY messages from the server correctly. It seems to just ignore them leading to queries hanging until timeout.

An additional issue is that the RWLock we use for reconnecting does not prefer writers, so new queries, which acquire a read lock, can starve the writer, meaning that a reconnect never happens.

This mostly affects the wallet-proxy and middleware. The client itself makes a few queries per connection so it is not affected in practice.

**Steps to Reproduce**

Connect the middleware to a server that sends a GOAWAY frame after some number of queries.
Make simple queries to the middleware.

**Expected Result**

Queries keep working, the middleware reconnects immediately after it notices a disconnect.

**Actual Result**

The middleware tends to take many minutes to reconnect if the number of queries is high enough.

**Versions**
- Software Version: concordium-client all versions, wallet-proxy all-versions middleware all-versions

abizjak:
There are two main changes
- use an RWLock that prefers writers, so that new queries do not prevent a reconnect
- Detect GOAWAY frames more directly instead of waiting for timeout.

## Purpose

Fixes #188 

## Changes

- Use the RWLock we use in the node, which strongly prefers writers.
- Add custom handling of GoAway frames and reconnect immediately.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/src/Concordium/Client/GRPC.hs
+++ b/src/Concordium/Client/GRPC.hs
@@ -484,7 +484,7 @@ acquireRead RWLock{..} = mask_ go
         go
 
 -- |Acquire a write lock. This will block when there are active readers or
--- writers. When this is operation is blocked it also blocks new readers from
+-- writers. When this operation is blocked it also blocks new readers from
 -- acquiring the lock.
 acquireWrite :: RWLock -> IO ()
 acquireWrite RWLock{..} = mask_ $ go False
--- a/src/Concordium/Client/GRPC.hs
+++ b/src/Concordium/Client/GRPC.hs
@@ -24,7 +24,6 @@ import           Concordium.Types as Types
 import           Concordium.ID.Types as IDTypes
 
 import           Control.Exception
-import qualified Control.Concurrent.ReadWriteLock as RW
 import           Control.Concurrent.Async
 import           Control.Concurrent
 import           Control.Monad.Fail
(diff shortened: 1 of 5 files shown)


src/Concordium/Client/GRPC.hs
```diff
@@ -334,6 +354,200 @@ getCryptographicParameters hash = withUnary (call @"getCryptographicParameters")
...
+            -- Due to fairness of MVars next time another thread will make progress
+            -- so we are going to end up after a finite number of iterations, in a WriteLocked state.
+            putMVar rwlState st
+            -- Since this branch seems to be compiled into a loop without
+            -- allocations by GHC with -O2 we need to explicitly yield to allow others
+            -- to make progress. Otherwise with sufficient contention this loop ends up
+            -- starving other threads since they are never scheduled. This then also means
+            -- the loop never terminates since no other thread transitions from the Free
+            -- to WriteLocked state.
+            yield
```
EmilLa1, line 467: 👍  :)

src/Concordium/Client/GRPC.hs
```diff
@@ -334,6 +354,200 @@ getCryptographicParameters hash = withUnary (call @"getCryptographicParameters")
...
+              readMVar rwlReadLock
+              go
+      lockState@(WriteLocked _) -> do
+        -- There is an active writer. Do nothing and wait until the writer is done.
+        putMVar rwlState lockState
+        readMVar rwlWriteLock
+        go
+
+-- |Acquire a write lock. This will block when there are active readers or
+-- writers. When this is operation is blocked it also blocks new readers from
```
EmilLa1, line 487:
```suggestion
-- writers. When this operation is blocked it also blocks new readers from
```

EmilLa1 (approved):

PR #191, opened 2022-11-15
Deserialize parameters in `printTransactionStatus`.

vikt0r0:
## Purpose

Deserialize receive function parameters in output of `printTransactionStatus` to JSON when receive function schema is present, as currently only hexadecimal is printed.

## Changes

- Include relevant `ContractInfo` values for each transaction event in `printTransactionStatus` instead of just event schema info.
- Implement functionality for extracting the relevant schema information from `ContractInfo` for an event.
- Implement parameter parsing according to receive function schema, i.e. in the case for case `Concordium.Types.Updated` in `printTransactionStatus`.
- Change existing parsing methods of contract events to use schema information from `ContractInfo` instead.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/src/Concordium/Client/Output.hs
+++ b/src/Concordium/Client/Output.hs
@@ -545,9 +545,9 @@ parseTransactionBlockResult status =
 -- the events.
 printTransactionStatus :: TransactionStatusResult
                        -> Bool
-                       -> Maybe (Map.Map Types.BlockHash [(Types.Event, Maybe CS.SchemaType)])
+                       -> Maybe (Map.Map Types.BlockHash [(Types.Event, Maybe CI.ContractInfo)])
                        -> Printer
-printTransactionStatus status verbose eventsAndSchemas =
+printTransactionStatus status verbose contrInfoWithSchemas =
   case tsrState status of
     Received -> tell ["Transaction is pending."]
     Absent -> tell ["Transaction is absent."]
@@ -560,7 +560,7 @@ printTransactionStatus status verbose eventsAndSchemas =
                  "Transaction is committed into block %s with %s."
                  (show hash)
                  (showOutcomeFragment outcome)]
-          tell $ showOutcomeResult verbose (lookupEventsAndSchemas hash) $ Types.tsResult outcome
+          tell $ showOutcomeResult verbose (lookupContrInfo hash) $ Types.tsResult outcome
         MultipleBlocksUnambiguous hashes outcome -> do
           tell [printf
                  "Transaction is committed into %d blocks with %s:"
@@ -568,7 +568,7 @@ printTransactionStatus status verbose eventsAndSchemas =
                  (showOutcomeFragment outcome)]
           tell $ hashes <&> printf "- %s" . show
           case hashes of
-            hash:_ -> tell $ showOutcomeResult verbose (lookupEventsAndSchemas hash) $ Types.tsResult outcome
+            hash:_ -> tell $ showOutcomeResult verbose (lookupContrInfo hash) $ Types.tsResult outcome
             _ -> return ()
         MultipleBlocksAmbiguous blocks -> do
           tell [printf
@@ -578,7 +578,7 @@ printTransactionStatus status verbose eventsAndSchemas =
             tell [ printf "- %s with %s:"
                      (show hash)
                      (showOutcomeFragment outcome) ]
-            tell $ showOutcomeResult True (lookupEventsAndSchemas hash) (Types.tsResult outcome) <&> ("  * " ++)
+            tell $ showOutcomeResult True (lookupContrInfo hash) (Types.tsResult outcome) <&> ("  * " ++)
     Finalized ->
       case parseTransactionBlockResult status of
         NoBlocks ->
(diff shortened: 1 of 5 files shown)


vikt0r0:
Example output can be found here:

[example.txt](https://github.com/Concordium/concordium-client/files/10016778/example.txt)

src/Concordium/Client/Types/Contract/Info.hs
```diff
@@ -55,7 +60,7 @@ import Concordium.Client.Cli
 -- Logs fatally on internal errors that should never occur, namely:
 --  - Schema data has already been added.
 --  - The version of module schema and contract info does not match.
-addSchemaData :: ContractInfo -> CS.ModuleSchema -> ClientMonad IO (Maybe ContractInfo)
+addSchemaData :: (MonadIO m) => ContractInfo -> CS.ModuleSchema -> ClientMonad m (Maybe ContractInfo)
```
vikt0r0, line 63: Mostly out of curiosity. Should this function ideally not be pure? I don't see it doing any client or IO interaction.

abizjak: It does do logging to stdout/err. Or killing the process.

So it is effectful. I think it is perfectly sensible to be how it is. THis is ulimately about user interaction.

src/Concordium/Client/Runner.hs
```diff
@@ -545,60 +545,62 @@ checkAndGetMemo memo pv = do
...
-getTxEventsAndSchemas schemaFile status = do
+
+-- |Get `ContractInfo` for all events in all blocks in which a transaction is present.
+-- Returns a map from blockhashes of blocks in which the transaction is present to the
+-- events of the transaction in that block. Each event figures in a pair with an optional
+-- `ContractInfo` value containing contract schema info associated with each event of the
+-- transaction in that block.
+-- Optionally takes a path to a schema file to be parsed and returned. The schema contained
+-- in the file will then take precedence over schemas embedded in the module of the and will
+-- thus be included in the `ContractInfo` for all events.
```
abizjak, line 592: Something is a bit off in this sentence.

src/Concordium/Client/Runner.hs
```diff
@@ -545,60 +545,62 @@ checkAndGetMemo memo pv = do
...
--- be included in its place for all events.
-getTxEventsAndSchemas :: (MonadIO m, MonadFail m)
-               => Maybe FilePath
-               -> TransactionStatusResult
-               -> ClientMonad m (Map.Map Types.BlockHash [(Types.Event, Maybe CS.SchemaType)])
-getTxEventsAndSchemas schemaFile status = do
+
+-- |Get `ContractInfo` for all events in all blocks in which a transaction is present.
+-- Returns a map from blockhashes of blocks in which the transaction is present to the
+-- events of the transaction in that block. Each event figures in a pair with an optional
```
abizjak, line 587: Figures?

abizjak (changes requested):
I think this looks almost there. I think there is one issue with reordering though that needs to be addressed.

src/Concordium/Client/Output.hs
```diff
@@ -787,27 +793,48 @@ showEvent verbose stM = \case
...
+
         -- Show a string representation of the contract event.
-        eventToString :: Wasm.ContractEvent -> String
-        eventToString e = case toJSON' e of
-          Nothing -> [i|Event(raw): #{show e}|]
-          Just json -> [i|Event(parsed):\n#{indentBy 4 json}|] 
+        eventToString :: Wasm.ContractEvent -> Either String String
+        eventToString ce@(Wasm.ContractEvent bs) = case toJSON' eventSchemaM bs of
+          Nothing -> Left [i|Event (raw): '#{ce}'|]
+          Just json -> Right [i|Event:\n#{indentBy 4 $ "'" <> json <> "'"}|]
```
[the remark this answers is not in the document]
vikt0r0: This is to count the number of events that could be succesfully parsed using the schema:

`811:    else [i|, of which #{length $ filter isLeft $ map eventToString evs} were succesfully parsed. |])`

Do you believe that a boolean is better choice in this case?

src/Concordium/Client/Output.hs
```diff
@@ -787,27 +793,48 @@ showEvent verbose stM = \case
     showDelegationTarget Types.DelegatePassive = "Passive delegation"
     showDelegationTarget (Types.DelegateToBaker bid) = "Baker ID " ++ show bid
 
+    -- Attempt to decode a bytestring according to a schema.
+    -- If there is no schema, or decoding fails @Nothing@ is returned.
```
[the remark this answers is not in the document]
vikt0r0: It's just a helper for `eventToString` and `parameterToString` which conveniently takes a schema type wrapped in Maybe. Since we do not intend to use the error message, the semantics are that **in case of a decoding failure** or **if a schema is not present**, Nothing is returned.

Should I modify it to return the error? In that case, then I believe the signature should probably be:

`toJSON' :: CS.SchemaType -> BSS.ShortByteString -> Either String String`

or

`toJSON' :: Maybe CS.SchemaType -> BSS.ShortByteString -> Either String String`

I am not sure whether this adds any benefits, or makes anything simpler?

src/Concordium/Client/Output.hs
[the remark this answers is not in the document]
abizjak: I'm fine either way. Whatever is simpler. But please add a comment on what Left/Right means if you keep it.

src/Concordium/Client/Output.hs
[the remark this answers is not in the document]
abizjak: Ok.If it is going to be ignored then we don't have to keep it. But maybe add what you wrote above (about the error not being useful) to the documentation.

src/Concordium/Client/Runner.hs
vikt0r0: Sorry, this is Dan(ish-en)glish. I changed it.

abizjak (approved):

PR #192, opened 2022-11-16
Reconnect also when retryNum=1 and a GOAWAY frame is received.

abizjak:
## Purpose

Make it so that the client reconnects on the first attempt after handling the GOAWAY reset.
This makes the "retryNum" setting correct also in this scenario.

## Changes

Clear the MVar when closing the connection.

## Checklist

      hard-to-understand areas.


EmilLa1 (approved):
I haven’t tested it, but code and doc seems reasonable.

abizjak:
I have been testing it this evening (via the wallet-proxy queries) and it seems to work as intended.

It worked previously as well as long as retryNum was at least 1, but that is not the default :)

PR #202, opened 2023-01-11
Migrate client to use GRPC V2

vikt0r0:
### Purpose

### Changes
- [x] Add bootstrapping to generate code from the V2 API protocol buffer schema (in `Setup.hs`).
- [x] Migrate GRPC call helpers to use the V2 GRPC API:
- [ ] Add functional "inverse" to the `ToProto` class (i.e. `FromProto`), allowing for conversion between datatypes generated from protocol buffer schemas and their corresponding datatype in (e.g. those in `Concordium.Types.*`, `Concordium.Common`) where applicable. These are not necessarily one-to-one, so checks should be added accordingly.
- [ ] (Maybe) refactor `ToProto` and `FromProto` class and instance declarations to `concordium-base`.
- [ ] Implement V2 GRPC API calls.
- [ ] Rewrite/refactor client functions and helpers to use V2 GRPC API
- [ ] Add tests for V2 GRPC API calls


src/Concordium/Client/GRPC2.hs
```diff
@@ -0,0 +1,2676 @@
...
+                then go 0 msgs' -- reset wait time to 10ms.
+                else
+                    if res == -1
+                        then do
+                            let delay = (2 ^ n) * 10_000
+                            threadDelay delay
+                            go (min (n + 1) (10 :: Int)) msgs -- maximum delay is 10 seconds
+                        else return () -- the sender channel is now dropped, so we stop.
+
+-- V2 stuff starts here:
```
vikt0r0, line 2140: @abizjak The changes start here, above is just lifted (temporarily) from GRPC2.hs of the client.

src/Concordium/Client/GRPC2.hs
```diff
@@ -0,0 +1,3222 @@
+{-# LANGUAGE DataKinds #-}
```
abizjak, line 1: If we are copying this from the node we should do that in a separate PR so that we can also update the node, and have a clear separation so we don't introduce issues by accident.

src/Concordium/Client/GRPC2.hs
```diff
@@ -0,0 +1,3222 @@
...
+getInstanceInfoV2 cAddress blockHash = withUnaryCoreV2 (callV2 @"getInstanceInfo") msg (fmap fromProto <$>)
+  where
+    msg = defMessage & ProtoFields.blockHash .~ toProto blockHash & ProtoFields.address .~ toProto cAddress
+
+getNextSequenceNumberV2 :: (MonadIO m) => AccountAddress -> ClientMonad m (GRPCResult QueryTypes.NextAccountNonce)
+getNextSequenceNumberV2 accAddress = withUnaryCoreV2 (callV2 @"getNextAccountSequenceNumber") msg (fmap fromProto <$>)
+  where
+    msg = toProto accAddress
+
+{- FIXME: Should we still use GlobalContext for these internally?
```
abizjak, line 3017: Yes. One is parsed, the protobuf just has byte arrays everywhere.

src/Concordium/Client/GRPC2.hs
```diff
@@ -0,0 +1,3222 @@
...
+getPassiveDelegationInfoV2 :: (MonadIO m) => BlockHashInput -> ClientMonad m (GRPCResult QueryTypes.PoolStatus)
+getPassiveDelegationInfoV2 blockHash = withUnaryCoreV2 (callV2 @"getPassiveDelegationInfo") msg (fmap fromProto <$>)
+  where
+    msg = toProto blockHash
+
+getPoolInfoV2 :: (MonadIO m) => BlockHashInput -> BakerId -> ClientMonad m (GRPCResult QueryTypes.PoolStatus)
+getPoolInfoV2 blockHash baker = withUnaryCoreV2 (callV2 @"getPoolInfo") msg (fmap fromProto <$>)
+  where
+    msg = defMessage & ProtoFields.blockHash .~ toProto blockHash & ProtoFields.baker .~ toProto baker
+
```
abizjak, line 2981: This looks good.

src/Concordium/Client/GRPC2.hs
```diff
@@ -0,0 +1,3222 @@
...
+    type Output' Proto.ContractTraceElement = Event
+    fromProto cte =
+        case ( cte ^? ProtoFields.updated
+             , cte ^? ProtoFields.transferred
+             , cte ^? ProtoFields.interrupted
+             , cte ^? ProtoFields.resumed
+             , cte ^? ProtoFields.upgraded
+             ) of
+            (Just updated, Nothing, Nothing, Nothing, Nothing) ->
+                Updated
```
abizjak, line 2801: As we discussed, this is not the way to go about it. For this one the return would be `Maybe Event` and then

```
fromProto cte = do
    inner <- cte ^. Proto.maybe'element
    case inner of 
        Proto.ContractTraceElement'Element ue -> fromProot ue
        ...
```

src/Concordium/Client/GRPC2.hs
```diff
@@ -0,0 +1,3222 @@
...
+
+data InvokeInstanceInput = InvokeInstanceInput
+    { iiBlockHash :: BlockHashInput
+    , iiInvoker :: Maybe Address
+    , iiInstance :: ContractAddress
+    , iiAmount :: Amount
+    , iiEntrypoint :: Wasm.ReceiveName
+    , iiParameter :: Wasm.Parameter
+    , iiEnergy :: Energy
+    }
```
abizjak, line 2758: This already exists as `data ContractContext = ContractContext`

src/Concordium/Client/GRPC2.hs
```diff
@@ -0,0 +1,3222 @@
...
+instance FromProto Proto.ProtocolVersion where
+    type Output' Proto.ProtocolVersion = ProtocolVersion
+    fromProto = \case
+        Proto.PROTOCOL_VERSION_1 -> P1
+        Proto.PROTOCOL_VERSION_2 -> P2
+        Proto.PROTOCOL_VERSION_3 -> P3
+        Proto.PROTOCOL_VERSION_4 -> P4
+        Proto.PROTOCOL_VERSION_5 -> P5
+        Proto.PROTOCOL_VERSION_6 -> P6
+        Proto.ProtocolVersion'Unrecognized _ -> error "ouch"
```
abizjak, line 2538: We need to do proper error handling and return Option/Result's

src/Concordium/Client/GRPC2.hs
```diff
@@ -0,0 +1,3222 @@
...
+
+instance FromProto Proto.BlockHash where
+    type Output' Proto.BlockHash = BlockHash
+    fromProto bh = BlockHash $ deMkSerialize bh
+
+instance FromProto Proto.ModuleRef where
+    type Output' Proto.ModuleRef = ModuleRef
+    fromProto mr =
+        ModuleRef
+            { moduleRef = Hash $ FBS.fromByteString (mr ^. ProtoFields.value)
```
abizjak, line 2474: The length is not necessarily correct. So you need to validate.

src/Concordium/Client/GRPC2.hs
```diff
@@ -0,0 +1,3222 @@
...
+            , aiAccountCredentials =
+                Map.fromAscList
+                    . fmap (\(k, v) -> (CredentialIndex $ fromIntegral k, Versioned 0 $ fromProto v))
+                    . Map.toAscList
+                    $ ai ^. ProtoFields.creds -- FIXME: Specify version?
+            , aiAccountThreshold = fromProto $ ai ^. ProtoFields.threshold
+            , aiAccountEncryptedAmount = fromProto $ ai ^. ProtoFields.encryptedBalance
+            , aiAccountEncryptionKey = fromProto $ ai ^. ProtoFields.encryptionKey
+            , aiAccountIndex = fromProto $ ai ^. ProtoFields.index
+            , aiStakingInfo = fromProto . fromJust $ ai ^. ProtoFields.maybe'stake -- FIXME: Do not use fromJust
```
abizjak, line 2455: Indeed, to not :)

src/Concordium/Client/GRPC2.hs
```diff
@@ -0,0 +1,3222 @@
...
+            , pValidTo = fromProto $ p ^. ProtoFields.validTo
+            , pItems = deMkAttributes $ p ^. ProtoFields.attributes
+            }
+      where
+        deMkAttributes =
+            Map.fromAscList
+                . fmap
+                    ( \(tag, value) ->
+                        ( AttributeTag $ fromIntegral tag
+                        , AttributeValue $ fromRight (error "oops") (S.runGet (S.getShortByteString 0) value) -- FIXME: Is n == 0 ok here? Do not use fromRight.
```
abizjak, line 2413: Do not use fromRight.  I don't understand this parsing. The attribute value is a byte array. Why are you parsing 0 bytes?

src/Concordium/Client/GRPC2.hs
```diff
@@ -0,0 +1,3222 @@
...
+
+instance FromProto Proto.ArThreshold where
+    type Output' Proto.ArThreshold = Threshold
+    fromProto = Threshold . deMkWord8
+
+instance FromProto (Map.Map Word32 Proto.ChainArData) where
+    type Output' (Map.Map Word32 Proto.ChainArData) = Map.Map ArIdentity ChainArData
+    fromProto = Map.fromAscList . fmap (bimap ArIdentity protoToData) . Map.toAscList
+      where
+        protoToData d = fromRight (error "oops") (S.decode $ d ^. ProtoFields.encIdCredPubShare)
```
abizjak, line 2344: Do not raise errors.

src/Concordium/Client/GRPC2.hs
```diff
@@ -0,0 +1,3222 @@
...
+    type Output' Proto.BakerAggregationVerifyKey = BakerAggregationVerifyKey
+    fromProto = deMkSerialize
+
+instance FromProto Proto.OpenStatus where
+    type Output' Proto.OpenStatus = OpenStatus
+    fromProto = \case
+        Proto.OPEN_STATUS_OPEN_FOR_ALL -> OpenForAll
+        Proto.OPEN_STATUS_CLOSED_FOR_NEW -> ClosedForNew
+        Proto.OPEN_STATUS_CLOSED_FOR_ALL -> ClosedForAll
+        _ -> error "oops"
```
abizjak, line 2288: Return `Maybe` intead.

src/Concordium/Client/GRPC2.hs
```diff
@@ -0,0 +1,3222 @@
...
+instance FromProto Proto.DelegationTarget where
+    type Output' Proto.DelegationTarget = DelegationTarget
+    fromProto dt = case (dt ^? ProtoFields.passive, dt ^? ProtoFields.baker) of
+        (Just _, _) -> DelegatePassive
+        (_, Just baker) -> DelegateToBaker $ fromProto baker
+        _ -> error "Oops"
+
+instance FromProto Proto.StakePendingChange where
+    type Output' Proto.StakePendingChange = (StakePendingChange' Timestamp, StakePendingChange' UTCTime)
+    fromProto spc = case (spc ^? ProtoFields.reduce, spc ^? ProtoFields.remove) of
```
abizjak, line 2251: Please do proper pattern matching. Not this way.

src/Concordium/Client/GRPC2.hs
```diff
@@ -0,0 +1,3222 @@
...
+instance FromProto Proto.BakerId where
+    type Output' Proto.BakerId = BakerId
+    fromProto = BakerId . AccountIndex . deMkWord64
+
+instance FromProto Proto.DelegationTarget where
+    type Output' Proto.DelegationTarget = DelegationTarget
+    fromProto dt = case (dt ^? ProtoFields.passive, dt ^? ProtoFields.baker) of
+        (Just _, _) -> DelegatePassive
+        (_, Just baker) -> DelegateToBaker $ fromProto baker
+        _ -> error "Oops"
```
abizjak, line 2247: No errors.

src/Concordium/Client/GRPC2.hs
```diff
@@ -0,0 +1,3222 @@
...
+    fromProto eb =
+        AccountEncryptedAmount
+            { _selfAmount = fromProto $ eb ^. ProtoFields.selfAmount
+            , _startIndex = EncryptedAmountAggIndex . fromIntegral $ eb ^. ProtoFields.startIndex
+            , _aggregatedAmount =
+                case (fromProto <$> eb ^? ProtoFields.aggregatedAmount, fromIntegral <$> eb ^? ProtoFields.numAggregated) of
+                    (Just aggAmt, Just numAgg) -> Just (aggAmt, numAgg)
+                    (Nothing, Nothing) -> Nothing
+                    _ -> error "oops"
+            , _incomingEncryptedAmounts = Empty -- FIXME: Field is not included in other direction. Should it be?
```
abizjak, line 2231: What do you mean it is not included in the other direction? Most accounts will have this empty, so you need to find one which has some incoming encrypted transfers. Just make one using the client to have an account to test with.

src/Concordium/Client/GRPC2.hs
```diff
@@ -0,0 +1,3222 @@
...
+                        then do
+                            let delay = (2 ^ n) * 10_000
+                            threadDelay delay
+                            go (min (n + 1) (10 :: Int)) msgs -- maximum delay is 10 seconds
+                        else return () -- the sender channel is now dropped, so we stop.
+
+-- V2 stuff starts here:
+
+-- |A block hash input
+data BlockHashInput = Best | LastFinal | Given BlockHash
```
abizjak, line 2150: Strict fields please.

src/Concordium/Client/GRPC2.hs
```diff
@@ -0,0 +1,3222 @@
...
+instance FromProto Proto.AccountAddress where
+    type Output' Proto.AccountAddress = AccountAddress
+    fromProto = deMkSerialize
+
+instance FromProto Proto.CredentialRegistrationId where
+    type Output' Proto.CredentialRegistrationId = RawCredentialRegistrationID
+    fromProto = deMkSerialize
+instance FromProto Proto.AccountIdentifierInput where
+    type Output' Proto.AccountIdentifierInput = AccountIdentifierInput
+    fromProto a = case (a ^? ProtoFields.credId, a ^? ProtoFields.address, a ^? ProtoFields.accountIndex) of
```
abizjak, line 2167: Please use pattern matching on the discriminant. Not this. This is hard to maintain.

src/Concordium/Client/GRPC2.hs
```diff
@@ -0,0 +1,3222 @@
...
+            res <- callback (castPtr headPtr) (fromIntegral len)
+            if res == 0
+                then go 0 msgs' -- reset wait time to 10ms.
+                else
+                    if res == -1
+                        then do
+                            let delay = (2 ^ n) * 10_000
+                            threadDelay delay
+                            go (min (n + 1) (10 :: Int)) msgs -- maximum delay is 10 seconds
+                        else return () -- the sender channel is now dropped, so we stop.
```
abizjak, line 2145: This is definitely not needed here. It is very node specific due to the FFI there.

src/Concordium/Client/GRPC2.hs
```diff
@@ -0,0 +1,3222 @@
...
+-- |Like 'deMkWord64' above, but for Word16 fields instead.
+deMkWord16 ::
+    ( Data.ProtoLens.Field.HasField
+        b
+        "value"
+        Word32
+    ) =>
+    b ->
+    Word16
+deMkWord16 val = fromIntegral $ shiftR (shiftL (val ^. ProtoFields.value) 16) 16
```
abizjak, line 247: What exactly is happening here? You cannot do this. You have to validate that it fits into Word16.

src/Concordium/Client/GRPC2.hs
```diff
@@ -0,0 +1,3222 @@
...
+-- |Like 'mkWord32', but for Word8 fields instead.
+deMkWord8 ::
+    ( Data.ProtoLens.Field.HasField
+        b
+        "value"
+        Word32
+    ) =>
+    b ->
+    Word8
+deMkWord8 val = fromIntegral $ shiftR (shiftL (val ^. ProtoFields.value) 24) 24
```
abizjak, line 258: Same here.

abizjak (commented):
Good start.

I did not comment on all the FromProto instances since I assume that you are already handling the changes based on the discussions we had on slack.

vikt0r0:
@abizjak Thanks for the review! Is it fine to use Maybe for the return values, or do you believe it may be better to use Either with some error message, e.g. something like:

"could not do <someTypeOfAction> for field <someField>: <someMoreSpecificErrorMessageHere>"?

abizjak:
> @abizjak Thanks for the review! Is it fine to use Maybe for the return values, or do you believe it may be better to use Either with some error message, e.g. something like:
> 
> ```
> could not do <someTypeOfAction> for field <someField>: <someMoreSpecificErrorMessageHere>
> ```

These errors should not leak to the users unless there are bugs. However in principle it is always better to use `Either String` rather than `Maybe`. It is easier to debug. But it does mean considerably more time to write useful messages.

So my preference would be to use Either String, but only if this does not delay things too much.

src/Concordium/Client/GRPC2.hs
vikt0r0: I agree, my question on slack was regarding this.

Could you write a quick example of what that would look like for the above case? I can not figure this out, and the documentation is a bit sparse IMO.

abizjak:
```haskell
instance FromProto Proto.AccountIdentifierInput where
    type Output' Proto.AccountIdentifierInput = Maybe AccountIdentifierInput
    fromProto a = do
      aii <- a ^. ProtoFields.maybe'accountIdentifierInput
      case aii of
        Proto.AccountIdentifierInput'Address addr -> AccAddress <$> fromProto addr
        Proto.AccountIdentifierInput'CredId cid -> CredRegID <$> fromProto cid
        Proto.AccountIdentifierInput'AccountIndex ai -> Just (fromProto ai)
```

assuming you've changed `FromProto AccountAddress` to return `Maybe AccountAddress`.

abizjak: If LSP is of no help you can see the generated file with all the types in something like

```
.stack-work/dist/x86_64-linux-tinfo6/Cabal-3.6.3.0/build/autogen/Proto/V2/Concordium/Types.hs
```

It has somewhat readable documentation, for example
```haskell
{- | Fields :
     
         * 'Proto.V2.Concordium.Types_Fields.maybe'accountIdentifierInput' @:: Lens' AccountIdentifierInput (Prelude.Maybe AccountIdentifierInput'AccountIdentifierInput)@
         * 'Proto.V2.Concordium.Types_Fields.maybe'address' @:: Lens' AccountIdentifierInput (Prelude.Maybe AccountAddress)@
         * 'Proto.V2.Concordium.Types_Fields.address' @:: Lens' AccountIdentifierInput AccountAddress@
         * 'Proto.V2.Concordium.Types_Fields.maybe'credId' @:: Lens' AccountIdentifierInput (Prelude.Maybe CredentialRegistrationId)@
         * 'Proto.V2.Concordium.Types_Fields.credId' @:: Lens' AccountIdentifierInput CredentialRegistrationId@
         * 'Proto.V2.Concordium.Types_Fields.maybe'accountIndex' @:: Lens' AccountIdentifierInput (Prelude.Maybe AccountIndex)@
         * 'Proto.V2.Concordium.Types_Fields.accountIndex' @:: Lens' AccountIdentifierInput AccountIndex@ -}
data AccountIdentifierInput
  = AccountIdentifierInput'_constructor {_AccountIdentifierInput'accountIdentifierInput :: !(Prelude.Maybe AccountIdentifierInput'AccountIdentifierInput),
                                         _AccountIdentifierInput'_unknownFields :: !Data.ProtoLens.FieldSet}
  deriving stock (Prelude.Eq, Prelude.Ord)
instance Prelude.Show AccountIdentifierInput where
  showsPrec _ __x __s
    = Prelude.showChar
        '{'
        (Prelude.showString
           (Data.ProtoLens.showMessageShort __x) (Prelude.showChar '}' __s))
data AccountIdentifierInput'AccountIdentifierInput
  = AccountIdentifierInput'Address !AccountAddress |
    AccountIdentifierInput'CredId !CredentialRegistrationId |
    AccountIdentifierInput'AccountIndex !AccountIndex
  deriving stock (Prelude.Show, Prelude.Eq, Prelude.Ord)

```

vikt0r0: Worked like a charm!

src/Concordium/Client/GRPC2.hs
vikt0r0: It is a little different. `ContractContext` has `Maybe BlockHash` rather than `BlockHashInput` . The protocol buffer schema also specifies a `BlockHashInput` in `InvokeInstanceRequest` which this type is modeled after.

If we use `ContractContext`, should this field simply not be set in the protocol buffer message when it is `Nothing`?

src/Concordium/Client/GRPC2.hs
vikt0r0: The documentation for the function specifies that this fails if fewer than `n` bytes are in the input? We have `n=0`in this case.

As I state, I know not to use `fromRight`, but utilizing `HasCallStack` has the added benefits of having the code compile before settling on an error handling strategy :-)

vikt0r0:
> > @abizjak Thanks for the review! Is it fine to use Maybe for the return values, or do you believe it may be better to use Either with some error message, e.g. something like:
> > ```
> > could not do <someTypeOfAction> for field <someField>: <someMoreSpecificErrorMessageHere>
> > ```
> 
> These errors should not leak to the users unless there are bugs. However in principle it is always better to use `Either String` rather than `Maybe`. It is easier to debug. But it does mean considerably more time to write useful messages.
> 
> So my preference would be to use Either String, but only if this does not delay things too much.

I prefer that too, but settled for Maybe. Since everything is in a monadic context, perhaps changing it should not be too cumbersome, should it eventually become relevant.

src/Concordium/Client/GRPC2.hs
vikt0r0: Okay. I ended up adding Rust methods and FFI bindings for instantiating the struct corresponding to the payload in base.

vikt0r0:
I "tested" getNodeInfoV2 and getInstanceInfoV2 just for sanity-checking; for both calls payloads are returned. For getNodeInfoV2 the payload is succesfully translated to the "native" Haskell data type:

```
NodeInfo {peerVersion = "5.1.3", localTime = 1674322373105, peerUptime = 26757, networkInfo = NetworkInfo {nodeId = "348b92750be71508", peerTotalSent = 724, peerTotalReceived = 957, avgBpsIn = 21461, avgBpsOut = 18904}, details = Just NodePassive}
```

However for getInstanceInfoV2, the deserialization fails, with the payload looking something like:

```
{v1 { owner { value: "\321\200O\247W\002\354\204\242\030\341!\334\342\305\276ok\n\n%\261x\362\231\034\001\233\027\227\325\036" } amount { } methods { value: "weather.get" } methods { value: "weather.set" } name { value: "init_weather" } source_module { value: "g\325hC;\327.C&$\037&\"\023\327\177Dm\270\272\003\337\2725\032\343\\\033.~Q\t" } }}
```

This shows that while the approach used here is viable, input needs to be validated (similarly to in the Rust SDK (?)) and  thorough testing probably need to be done.

abizjak:
> This shows that while the approach used here is viable, input needs to be validated (similarly to in the Rust SDK (?)) and testing needs to be done. Maybe we should arrange for a meeting with @mh-concordium for his valuable input :-)

I am not sure what you mean exactly. Of course we'll need to test it. That is what we also did when developing the Rust SDK (there is generally a small example for each endpoint there). Martin should get involved at the end for user testing, you need to do the testing of each endpoint before that.

What we discussed in the beginning, and the approach I wanted you to take is to do one endpoint at a time, and test it immediately. It makes for better development experience.

It is still not too late to do that. When you migrate the use of each command, e.g., `account show` you should test it immediately. Do not wait until the end. That will lead to problems since there are so many endpoints.

PR #207, opened 2023-01-18
New chain parameters

EmilLa1:
## Purpose

Support the new chain parameters

## Changes

Refactored code in places where needed to support the new chain parameters and how they are structured. 

Added a printing function for cpv2 which prints out the new 'consensus parameters' instead of the election difficulty param.

## Checklist

      hard-to-understand areas.


src/Concordium/Client/Output.hs
```diff
@@ -1088,7 +1090,58 @@ printChainParametersV1 ChainParameters {..} = tell [
...
+  [i|     * adding a credential deployment: #{_cpRewardParameters ^. gasAccountCreation}|],
+  [i|     * adding a chain update: #{_cpRewardParameters ^. gasChainUpdate}|],
+  "",
+  [i|\# Time parameters:|],
+  [i|  + reward period length: #{_cpTimeParameters ^. tpRewardPeriodLength} epochs|],
+  "",
+  [i|\# Consensus parameters: |],
+  [i|  + timeout parameters: #{_cpConsensusParameters ^. cpTimeoutParameters}|],
+  [i|  + minimum block time: #{_cpConsensusParameters ^. cpMinBlockTime}|],
+  [i|  + block energy limit: #{_cpConsensusParameters ^. cpBlockEnergyLimit}|],
```
abizjak, line 1143: Are you sure these display nicely? TimeoutParameter is a structure, no? So it won't exactly display nicely. Fix if correct.

abizjak (approved):
Please fix printing. Otherwise fine.

src/Concordium/Client/Output.hs
EmilLa1: Thanks for pointing out. Fixed.

PR #209, opened 2023-01-23
Implement non-streaming V2 GRPC API calls.

vikt0r0:
### Purpose

This PR concerns itself with implementing the non-streaming V2 GRPC node API calls. The changes implement the non-streaming API calls of the V2 GRPC API as well as functionality for translating between generated protocol buffer schema types and their native (Concordium) Haskell type equivalents. Specifically, this translation is used for interface parameters and return values.

### Depends on
https://github.com/Concordium/concordium-base/pull/307

### Changes
Implementation of helpers and API calls: 
  - **Implement (non-streaming) V2 GRPC API call helper(s)**
  - **Implement (non-streaming) V2 GRPC API calls**
    - shutdown
    - peerConnect
    - peerDisconnect
    - getBannedPeers
    - banPeer
    - unbanPeer
    - dumpStart
    - dumpStop
    - getAccountInfo
    - getModuleSource
    - getInstanceInfo
    - getNextAccountSequenceNumber
    - getNextUpdateSequenceNumbers
    - getConsensusInfo
    - getBlockInfo
    - getPoolInfo
    - getPassiveDelegationInfo
    - getBlocksAtHeight
    - getTokenomicsInfo
    - invokeInstance
    - getBranches
    - getElectionInfo
    - instanceStateLookup
    - getBlockItemStatus
    - getPeersInfo
    - getNodeInfo
    - sendBlockItem
    - getAccountTransactionSignHash
    - getBlockChainParameters
    - getBlockFinalizationSummary


vikt0r0:
To the reviewers; comments with the `FIXME`-prefix indicate lines where I have doubts about what approach to use.

src/Concordium/Client/GRPC2.hs
```diff
@@ -0,0 +1,2693 @@
...
+    , latency :: !Integer
+    }
+data PeerInfo = PeerInfo
+    { peerId :: !Text
+    , socketAddress :: !IpSocketAddress
+    , networkStats :: !NetworkStats
+    , consensusInfo :: !(Maybe PeerCatchupStatus)
+    }
+
+type PeersInfo = [PeerInfo]
```
abizjak, line 1959: Documentation.

src/Concordium/Client/GRPC2.hs
```diff
@@ -0,0 +1,2693 @@
...
+                ProtoFields.NodeInfo'BakerConsensusInfo'ActiveFinalizerCommitteeInfo' _ ->
+                    return ActiveFinalizerCommitteeInfo
+                ProtoFields.NodeInfo'BakerConsensusInfo'PassiveCommitteeInfo pcInfo ->
+                    PassiveBaker <$> fromProto pcInfo
+        return BakerConsensusInfo{..}
+
+data BakerConsensusInfo = BakerConsensusInfo
+    { bakerId :: !Integer
+    , status :: !BakerConsensusInfoStatus
+    }
```
abizjak, line 2043: Documentation.

src/Concordium/Client/GRPC2.hs
```diff
@@ -0,0 +1,2693 @@
...
+    , avgBpsIn :: !Integer
+    , avgBpsOut :: !Integer
+    }
+
+data NodeInfo = NodeInfo
+    { peerVersion :: !Text
+    , localTime :: !Timestamp
+    , peerUptime :: !Duration
+    , networkInfo :: !NetworkInfo
+    , details :: !(Maybe NodeDetails)
```
abizjak, line 2091: This is an odd choice here. Why not be explicit about being a bootstrapper or not?

src/Concordium/Client/GRPC2.hs
```diff
@@ -0,0 +1,2693 @@
...
+        avgBpsIn <- fromIntegral <$> nInfo ^? ProtoFields.avgBpsIn
+        avgBpsOut <- fromIntegral <$> nInfo ^? ProtoFields.avgBpsOut
+        return NetworkInfo{..}
+
+data NetworkInfo = NetworkInfo
+    { nodeId :: !Text
+    , peerTotalSent :: !Integer
+    , peerTotalReceived :: !Integer
+    , avgBpsIn :: !Integer
+    , avgBpsOut :: !Integer
```
abizjak, line 2083: Word64.

src/Concordium/Client/GRPC2.hs
```diff
@@ -0,0 +1,2693 @@
...
+        _cpExchangeRates <- do
+            euroPerEnergy <- fromProto =<< cParams ^? ProtoFields.euroPerEnergy
+            microCcdPerEuro <- fromProto =<< cParams ^? ProtoFields.microCcdPerEuro
+            return $ Parameters.makeExchangeRates euroPerEnergy microCcdPerEuro
+        _cpCooldownParameters <-
+            fmap Parameters.CooldownParametersV0
+                . fromProto
+                =<< cParams ^? ProtoFields.bakerCooldownEpochs
+        -- FIXME: I don't see this (required field) anywhere in the protobuf message.
+        _cpTimeParameters <- Nothing
```
abizjak, line 2107: There are no time parameters for chain parameters 0.

So set `_cpTimeParameters = TimeParametersV0`.

src/Concordium/Client/GRPC2.hs
```diff
@@ -0,0 +1,2693 @@
...
+            return Parameters.RewardParameters{..}
+        -- FIXME: ProtoFields.foundationAccount has type AccountAddress, but the type expects an AccountIndex.
+        _cpFoundationAccount <- Nothing
+        _cpPoolParameters <- do
+            fmap Parameters.PoolParametersV0 . fromProto =<< cParams ^? ProtoFields.minimumThresholdForBaking
+        {- FIXME: The following fields are present in the protobuf message for ChainParametersV0, but there
+                  are no corresponding record fields in the datatype for chain protocol 1.
+
+                    // Minimum threshold for becoming a baker.
+                    Amount minimum_threshold_for_baking = 10;
```
abizjak, line 2124:
```
data PoolParameters cpv where
    PoolParametersV0 ::
        { -- |Minimum threshold required for registering as a baker.
          _ppBakerStakeThreshold :: Amount
        } ->
        PoolParameters 'ChainParametersV0
```

src/Concordium/Client/GRPC2.hs
```diff
@@ -0,0 +1,2693 @@
...
+                    // Keys allowed to do root updates.
+                    HigherLevelKeys root_keys = 11;
+                    // Keys allowed to do level1 updates;
+                    HigherLevelKeys level1_keys = 12;
+                    // Keys allowed to do parameter updates.
+                    AuthorizationsV0 level2_keys = 13;
```
abizjak, line 2130: These are not part of chain parameters. It is part of `Updates`,
```haskell
-- |Current state of updatable parameters and update queues.
data Updates' (cpv :: ChainParametersVersion) = Updates
    { -- |Current update authorizations.
      _currentKeyCollection :: !(Hashed (UpdateKeysCollection cpv)),
      -- |Current protocol update.
      _currentProtocolUpdate :: !(Maybe ProtocolUpdate),
      -- |Current chain parameters.
      _currentParameters :: !(ChainParameters' cpv),
      -- |Pending updates.
      _pendingUpdates :: !(PendingUpdates cpv)
    }
    deriving (Show, Eq)
```
Specifically the update keys collection.

src/Concordium/Client/GRPC2.hs
```diff
@@ -0,0 +1,2693 @@
...
+data NodeInfo = NodeInfo
+    { peerVersion :: !Text
+    , localTime :: !Timestamp
+    , peerUptime :: !Duration
+    , networkInfo :: !NetworkInfo
+    , details :: !(Maybe NodeDetails)
+    }
+
+instance FromProto Proto.ChainParametersV0 where
+    type Output' Proto.ChainParametersV0 = Parameters.ChainParameters' 'ChainParametersV0
```
abizjak, line 2095: This is probably the wrong output type. The input type has extra data.

src/Concordium/Client/GRPC2.hs
```diff
@@ -0,0 +1,2693 @@
...
+        _cpRewardParameters <- do
+            _rpMintDistribution <- fromProto =<< cParams ^? ProtoFields.mintDistribution
+            -- \|Distribution of transaction fees.
+            _rpTransactionFeeDistribution <- fromProto =<< cParams ^? ProtoFields.transactionFeeDistribution
+            -- \|Rewards paid from the GAS account.
+            _rpGASRewards <- fromProto =<< cParams ^? ProtoFields.gasRewards
+            return Parameters.RewardParameters{..}
+        -- FIXME: ProtoFields.foundationAccount has type AccountAddress, but the type expects an AccountIndex.
+        _cpFoundationAccount <- Nothing
+        _cpPoolParameters <- Nothing
```
abizjak, line 2154: Also pool parameters are missing, and you are not making use of the keys.

src/Concordium/Client/GRPC2.hs
```diff
@@ -0,0 +1,2693 @@
...
+instance FromProto Proto.ChainParametersV1 where
+    type Output' Proto.ChainParametersV1 = Parameters.ChainParameters' 'ChainParametersV1
+    fromProto cParams = do
+        _cpElectionDifficulty <- fromProto =<< cParams ^? ProtoFields.electionDifficulty
+        _cpExchangeRates <- do
+            euroPerEnergy <- fromProto =<< cParams ^? ProtoFields.euroPerEnergy
+            microCcdPerEuro <- fromProto =<< cParams ^? ProtoFields.microCcdPerEuro
+            return $ Parameters.makeExchangeRates euroPerEnergy microCcdPerEuro
+        _cpCooldownParameters <- fromProto =<< cParams ^? ProtoFields.cooldownParameters
+        _cpTimeParameters <- Nothing -- FIXME: I do not see this (required field) anywhere in the protobuf message.
```
abizjak, line 2143: There is
```
  TimeParametersCpv1 time_parameters = 5;

```
on line 2858.

src/Concordium/Client/GRPC2.hs
```diff
@@ -0,0 +1,2693 @@
...
+    = ChainParameterOutputV0 !(Parameters.ChainParameters' 'ChainParametersV0)
+    | ChainParameterOutputV1 !(Parameters.ChainParameters' 'ChainParametersV1)
+
+instance FromProto Proto.ChainParameters where
+    type
+        -- FIXME: Perhaps there is some nice(r) way achieve this
+        --        with (constrained) type variables or type
+        --        families? But I am uncertain whether the class
+        --        has to change, or if this can be done "locally"?
+        --        Or does it not matter?
```
abizjak, line 2176: I don't quite see what you want here. You could instead return an existential type, but I don't think that is better. In fact I think it's worse.

src/Concordium/Client/GRPC2.hs
```diff
@@ -0,0 +1,2693 @@
...
+        { amount :: !Amount
+        , receiver :: !AccountAddress
+        }
+    | TransferWithMemoPayload
+        { amount :: !Amount
+        , receiver :: !AccountAddress
+        , memo :: !Memo
+        }
+    | RegisteredDataPayload
+        !RegisteredData
```
abizjak, line 2221: We don't need this. This is for cases when there is no SDK. We have the full transaction payload in Haskell.

src/Concordium/Client/GRPC2.hs
```diff
@@ -0,0 +1,2693 @@
...
+    { sender :: !AccountAddress
+    , sequenceNumber :: !Nonce
+    , energyAmount :: !Energy
+    , expiry :: !TransactionTime
+    }
+
+data PreAccountTransaction = PreAccountTransaction
+    { header :: !AccountTransactionHeader
+    , payload :: !AccountTransactionPayload
+    }
```
abizjak, line 2233: We don't need this. And we should not have it. We have transactions already.

src/Concordium/Client/GRPC2.hs
```diff
@@ -0,0 +1,2693 @@
...
+
+instance ToProto AccountTransactionPayload where
+    type Output AccountTransactionPayload = Proto.AccountTransactionPayload
+    toProto payload =
```
abizjak, line 2237: Remove.

src/Concordium/Client/GRPC2.hs
```diff
@@ -0,0 +1,2693 @@
...
+            ProtoFields.energyAmount .= toProto energyAmount
+            ProtoFields.expiry .= toProto expiry
+
+instance ToProto PreAccountTransaction where
+    type Output PreAccountTransaction = Proto.PreAccountTransaction
+    toProto PreAccountTransaction{..} =
+        Proto.make $ do
+            ProtoFields.header .= toProto header
+            ProtoFields.payload .= toProto payload
+
```
abizjak, line 2303: Don't need this.

src/Concordium/Client/GRPC2.hs
```diff
@@ -0,0 +1,2693 @@
...
+instance FromProto Proto.CryptographicParameters where
+    type Output' Proto.CryptographicParameters = CryptographicParameters
+    fromProto cParams =
+        do
+            genString <- cParams ^? ProtoFields.genesisString
+            bpGens <- cParams ^? ProtoFields.bulletproofGenerators
+            occKey <- cParams ^? ProtoFields.onChainCommitmentKey
+            createGlobalContext genString bpGens occKey
+
+getCryptographicParametersV2 :: (MonadIO m) => BlockHashInput -> ClientMonad m (GRPCResult (Maybe CryptographicParameters))
```
abizjak, line 2313: Documentation of all of the below functions.

abizjak (changes requested):

abizjak (changes requested):
Looking good overall. But there are quite a few small, but critical, issues.

src/Concordium/Client/GRPC2.hs
```diff
@@ -0,0 +1,2693 @@
...
+    type Output' Proto.NodeInfo'NetworkInfo = NetworkInfo
+    fromProto nInfo = do
+        nodeId <- nInfo ^? ProtoFields.nodeId . ProtoFields.value
+        peerTotalSent <- fromIntegral <$> nInfo ^? ProtoFields.peerTotalSent
+        peerTotalReceived <- fromIntegral <$> nInfo ^? ProtoFields.peerTotalReceived
+        avgBpsIn <- fromIntegral <$> nInfo ^? ProtoFields.avgBpsIn
+        avgBpsOut <- fromIntegral <$> nInfo ^? ProtoFields.avgBpsOut
+        return NetworkInfo{..}
+
+data NetworkInfo = NetworkInfo
```
Bargsteen, line 2078: Docs here and below

src/Concordium/Client/GRPC2.hs
```diff
@@ -0,0 +1,2693 @@
...
+        _cpCooldownParameters <-
+            fmap Parameters.CooldownParametersV0
+                . fromProto
+                =<< cParams ^? ProtoFields.bakerCooldownEpochs
+        -- FIXME: I don't see this (required field) anywhere in the protobuf message.
+        _cpTimeParameters <- Nothing
+        _cpAccountCreationLimit <- fromProto =<< cParams ^? ProtoFields.accountCreationLimit
+        _cpRewardParameters <- do
+            _rpMintDistribution <- fromProto =<< cParams ^? ProtoFields.mintDistribution
+            -- \|Distribution of transaction fees.
```
Bargsteen, line 2111: What is the meaning of `\|` here? Some sort of lambda comment? ;)

src/Concordium/Client/GRPC2.hs
```diff
@@ -0,0 +1,2693 @@
...
+        cp <- cParams ^. Proto.maybe'parameters
+        case cp of
+            Proto.ChainParameters'V0 v0 -> ChainParameterOutputV0 <$> fromProto v0
+            Proto.ChainParameters'V1 v1 -> ChainParameterOutputV1 <$> fromProto v1
+
+data VersionedModuleSource
+    = ModuleSourceV0 !ByteString
+    | ModuleSourceV1 !ByteString
+
+type AccountTransactionSignHash = ByteString
```
Bargsteen, line 2189: Docs

src/Concordium/Client/GRPC2.hs
Bargsteen: Some more context:
When you want to send a transaction, you need to sign the payload and include the signature. The issue is that it is the _serialized_ and hashed payload that you need to sign and the serialization must match the one used in the node. So if you create a transaction with the protobuf datatypes and serialize that, you will get a different payload (because protobuf has a different way of serializing the data). So you will end up signing the wrong payload, making the transaction invalid.
That is why we have the endpoint with `PreAccountTransaction` for getting the correct hash to sign for the transaction. 
But if you have data types for the transaction, which are serialized in the way the node expects, then you can get the right payload directly and sign it. That is what you should do here with the client.

src/Concordium/Client/GRPC2.hs
```diff
@@ -0,0 +1,2693 @@
...
+    toProto sbi = Proto.make $
+        case sbi of
+            AccountTransaction aTransaction ->
+                ProtoFields.accountTransaction .= toProto aTransaction
+            AccountCreation aCreation ->
+                ProtoFields.credentialDeployment .= toProto aCreation
+            UpdateInstruction uInstruction ->
+                ProtoFields.updateInstruction .= toProto uInstruction
+
+-- FIXME: Ergonomics: Should input, resp. output be ByteString?
```
Bargsteen, line 2372: As I understand it, the `Maybe` allows you to distinguish:
- an entry that exists, but has no data (empty bytestring)
- an entry that does _not_ eixst

So I think it is fine as is. But it needs docs ;)

Bargsteen (changes requested):
Nice work 😊

src/Concordium/Client/GRPC2.hs
```diff
@@ -0,0 +1,2693 @@
...
+import Data.ProtoLens.Field qualified as Field
+import Data.ProtoLens.Service.Types
+import Data.Sequence (Seq (Empty))
+import Data.String
+import Data.Time.Clock.POSIX (posixSecondsToUTCTime)
+import Proto.V2.Concordium.Service qualified as CS
+import Proto.V2.Concordium.Types qualified as ProtoFields
+import Proto.V2.Concordium.Types_Fields qualified as Proto
+
+{- |A helper function that serves as an inverse to `mkSerialize`,
```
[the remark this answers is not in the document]
vikt0r0: I thought that at some point, but it is the dual of its inverse (`mkSerialize`) which seems to be named after its use of `Proto.make`.

src/Concordium/Client/GRPC2.hs
```diff
@@ -0,0 +1,2693 @@
...
+    if v <= 255
+        then return $ fromIntegral v
+        else Nothing
+  where
+    v = val ^. ProtoFields.value
+
+{- |A helper class analogous to something like Aeson's FromJSON.
+ It exists to make it more manageable to convert the Protobuf types to
+ their internal Haskell type equivalents.
+-}
```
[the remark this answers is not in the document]
vikt0r0: Roger that. I just used it since this was used in the `ToProto` declaration.

src/Concordium/Client/GRPC2.hs
```diff
@@ -0,0 +1,2693 @@
...
+class FromProto a where
+    -- |The corresponding Haskell type.
+    type Output' a
```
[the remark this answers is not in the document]
vikt0r0: I think `Output` is fine, since it's consistent with `ToProto`.

src/Concordium/Client/GRPC2.hs
```diff
@@ -0,0 +1,2693 @@
...
+
+    -- |A conversion function from the protobuf type to its Haskell equivalent.
+    fromProto :: a -> Maybe (Output' a)
+
+fromProtoError :: Maybe a
```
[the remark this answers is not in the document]
vikt0r0: I added this at some point because I thought it might make it easier to change the (conversion error) wrapper monad eventually, but it seems I forgot to use it.

src/Concordium/Client/GRPC2.hs
```diff
@@ -0,0 +1,2693 @@
...
+            Proto.AccountIdentifierInput'AccountIndex accIdx -> AccIndex <$> fromProto accIdx
+            Proto.AccountIdentifierInput'Address addr -> AccAddress <$> fromProto addr
+            Proto.AccountIdentifierInput'CredId cId -> CredRegID . snd <$> fromProto cId
+
+instance ToProto AccountIdentifierInput where
+    type Output AccountIdentifierInput = Proto.AccountIdentifierInput
+    toProto = \case
+        CredRegID cred -> Proto.make $ ProtoFields.credId .= toProto cred
+        AccAddress addr -> Proto.make $ ProtoFields.address .= toProto addr
+        AccIndex accIdx -> Proto.make $ ProtoFields.accountIndex .= toProto accIdx
```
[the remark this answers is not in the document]
vikt0r0: It is needed for API input. It was added here at some point pending (the now finished) refactor of `ToProto` from the client to base.

src/Concordium/Client/GRPC2.hs
```diff
@@ -0,0 +1,2693 @@
...
+        releaseSchedule <- do
+            schedules <- ars ^? ProtoFields.schedules
+            mapM fromProto schedules
+        return AccountReleaseSummary{..}
+
+instance FromProto Proto.Timestamp where
+    type Output' Proto.Timestamp = (Timestamp, UTCTime)
+    fromProto t = do
+        time <- fromIntegral <$> t ^? ProtoFields.value
+        let utcTime = posixSecondsToUTCTime (fromIntegral time / 1_000.0)
```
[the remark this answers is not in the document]
vikt0r0: Yes, it seems so. `time` is the UNIX timestamp in milliseconds, so it has a precision of 10^-3 s whereas `NominalDiffTime` has precision of 10^-12 s.

src/Concordium/Client/GRPC2.hs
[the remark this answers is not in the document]
Bargsteen: Ok, so it will be moved eventually?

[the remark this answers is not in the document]
vikt0r0: It should be moved into the same modules of `base` in which `ToProto` is defined, since GHC will complain about orphan instances otherwise.

src/Concordium/Client/GRPC2.hs
```diff
@@ -0,0 +1,2693 @@
...
+instance FromProto Proto.DelegationTarget where
+    type Output' Proto.DelegationTarget = DelegationTarget
+    fromProto dTarget = do
+        dt <- dTarget ^. Proto.maybe'target
+        case dt of
+            Proto.DelegationTarget'Passive _ -> return DelegatePassive
+            Proto.DelegationTarget'Baker baker -> DelegateToBaker <$> fromProto baker
+
+instance FromProto Proto.StakePendingChange where
+    type Output' Proto.StakePendingChange = (StakePendingChange' Timestamp, StakePendingChange' UTCTime)
```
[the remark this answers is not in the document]
vikt0r0: So if I understand correctly, it is better to have:
```
    type Output' Proto.StakePendingChange = StakePendingChange' Timestamp
```
and some auxillary function
```
    convertStakePendingChangeToUtc :: StakePendingChange' Timestamp -> StakePendingChange' UTCTime
```

src/Concordium/Client/GRPC2.hs
```diff
@@ -0,0 +1,2693 @@
...
+    type Output' Proto.CredentialRegistrationId = (CredentialRegistrationID, RawCredentialRegistrationID)
+    fromProto crId = do
+        raw <- deMkSerialize crId
+        nonRaw <- credIdFromRaw raw
+        return (nonRaw, raw)
+      where
+        credIdFromRaw (RawCredentialRegistrationID fbs) =
+            case S.decode (FBS.toByteString fbs) of
+                Left _ -> Nothing
+                Right c -> c
```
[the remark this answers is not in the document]
vikt0r0: Do you want the client to fail with `error` here if the input is invalid? If that is the case, what's the point of wrapping the computation in a monad then?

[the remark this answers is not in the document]
abizjak: I am not sure what you are asking. I am just proposing you add a function `credIdFromRaw` to the same place where `unsafeCredIdfromRaw` is, instead of having multiple copies in the client project. That functioni should return Maybe CredId.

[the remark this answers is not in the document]
vikt0r0: I was just asking due to the ambiguity "this" and "here", but I completely agree that moving it to `Types.hs` is the right thing to do. Thanks.

src/Concordium/Client/GRPC2.hs
```diff
@@ -0,0 +1,2693 @@
...
+    fromProto eb = do
+        _selfAmount <- fromProto =<< eb ^? ProtoFields.selfAmount
+        _startIndex <- do
+            si <- eb ^? ProtoFields.startIndex
+            return $ EncryptedAmountAggIndex . fromIntegral $ si
+        _aggregatedAmount <- do
+            aa <- fromProto =<< eb ^? ProtoFields.aggregatedAmount
+            na <- eb ^? ProtoFields.numAggregated
+            return $ return (aa, na)
+        let _incomingEncryptedAmounts = Empty -- FIXME: Field is not included in other direction. Should it be?
```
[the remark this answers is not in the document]
abizjak: That is what I was saying above, if this is not the case it is a bug and needs to be fixed.

https://github.com/Concordium/concordium-base/pull/308/files

src/Concordium/Client/GRPC2.hs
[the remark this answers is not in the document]
abizjak: Yes, because that auxiliary function is just
```
convertStakePendingChangeToUtc = fmap timestampToUTCTime
```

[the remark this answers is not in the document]
vikt0r0: It's a functor! Lovely.

src/Concordium/Client/GRPC2.hs
```diff
@@ -0,0 +1,2693 @@
...
+instance FromProto Proto.OpenStatus where
+    type Output' Proto.OpenStatus = OpenStatus
+    fromProto = \case
+        Proto.OPEN_STATUS_OPEN_FOR_ALL ->
+            return OpenForAll
+        Proto.OPEN_STATUS_CLOSED_FOR_NEW ->
+            return ClosedForNew
+        Proto.OPEN_STATUS_CLOSED_FOR_ALL ->
+            return ClosedForAll
+        _ -> fromProtoError
```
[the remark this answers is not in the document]
vikt0r0: It most definitely can. Thanks for pointing that out.

src/Concordium/Client/GRPC2.hs
```diff
@@ -0,0 +1,2693 @@
...
+        pCreatedAt <- fromProto =<< p ^? ProtoFields.createdAt
+        pValidTo <- fromProto =<< p ^? ProtoFields.validTo
+        pItems <-
+            fmap Map.fromAscList
+                . mapM convert
+                . Map.toAscList
+                =<< p ^? ProtoFields.attributes
+        return Policy{..}
+      where
+        versionTag = 0 -- FIXME: Which version?
```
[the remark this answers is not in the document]
vikt0r0: It's used in line 481.

src/Concordium/Client/GRPC2.hs
```diff
@@ -0,0 +1,2693 @@
...
+        biBlockHeight <- fromProto =<< bi ^? ProtoFields.height
+        biBlockParent <- fromProto =<< bi ^? ProtoFields.parentBlock
+        biBlockLastFinalized <- fromProto =<< bi ^? ProtoFields.lastFinalizedBlock
+        biGenesisIndex <- fromProto =<< bi ^? ProtoFields.genesisIndex
+        biEraBlockHeight <- fromProto =<< bi ^? ProtoFields.eraBlockHeight
+        biBlockReceiveTime <- fmap snd $ fromProto =<< bi ^? ProtoFields.receiveTime
+        biBlockArriveTime <- fmap snd $ fromProto =<< bi ^? ProtoFields.arriveTime
+        biBlockSlot <- fromProto =<< bi ^? ProtoFields.slotNumber
+        biBlockSlotTime <- fmap snd $ fromProto =<< bi ^? ProtoFields.slotTime
+        biBlockBaker <- fmap fromProto =<< bi ^? ProtoFields.maybe'baker
```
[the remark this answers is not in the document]
vikt0r0: Yes. I will do this for optional fields moving on.

src/Concordium/Client/GRPC2.hs
```diff
@@ -0,0 +1,2693 @@
...
+            ProtoFields.relative
+                .= Proto.make
+                    ( do
+                        ProtoFields.genesisIndex .= toProto rGenesisIndex
+                        ProtoFields.height .= toProto rBlockHeight
+                        ProtoFields.restrict .= rRestrict
+                    )
+    toProto Absolute{..} =
+        Proto.make $
+            ProtoFields.absolute .= Proto.make (ProtoFields.height .= toProto aBlockHeight)
```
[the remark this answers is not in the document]
vikt0r0: This is used for input to `getBlocksAtHeightV2`.

src/Concordium/Client/GRPC2.hs
vikt0r0: Do you mean something like:

```
data NodeInfo =
    NodeNotRunning { ... }
  | NodePassive { ... }
  | NodeActive { bakerConsensusInfo :: !BakerConsensusInfo, ... }
```

I suppose `NodePassive` means it is a bootstrapper (rather than a baker) then?

src/Concordium/Client/GRPC2.hs
```diff
@@ -0,0 +1,2693 @@
...
+        updateSeqNumber <- fromProto =<< uiHeader ^? ProtoFields.sequenceNumber
+        updateEffectiveTime <- fromProto =<< uiHeader ^? ProtoFields.effectiveTime
+        updateTimeout <- fromProto =<< uiHeader ^? ProtoFields.timeout
+        let updatePayloadSize = PayloadSize 0 -- FIXME: What should go here, no such field in the protobuf equivalent.
+        return Updates.UpdateHeader{..}
+
+instance FromProto Proto.UpdateInstruction where
+    type Output' Proto.UpdateInstruction = Updates.UpdateInstruction
+    fromProto ui = do
+        uiSignHash <- Nothing -- FIXME: What should go here, no such field in the protobuf equivalent.
```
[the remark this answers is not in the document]
vikt0r0: `makeUpdateInstruction` has signature `RawUpdateInstruction -> Map.Map UpdateKeyIndex KeyPair -> UpdateInstruction`.

I'm not sure what to provide for this parameter. The payload has a field corresponding to `Map.Map UpdateKeyIndex Signature`. Should the `uiSignatures` of the `UpdateInstruction` be set manually to this?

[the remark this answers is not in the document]
abizjak: Right. Then we need to add a similar helper there that takes signatures instead of keypairs. Maybe it already exists in that module, cannot check now. The main point is that computing the hash should be done in the one place. Not here since it is tricky to make sure it is the same.

src/Concordium/Client/GRPC2.hs
```diff
@@ -0,0 +1,2693 @@
...
+    { iiBlockHash :: !BlockHashInput
+    , iiInvoker :: !Address
+    , iiInstance :: !ContractAddress
+    , iiAmount :: !Amount
+    , iiEntrypoint :: !Wasm.ReceiveName
+    , iiParameter :: !Wasm.Parameter
+    , iiEnergy :: !Energy
+    } -- FIXME: As mentioned by Aleš, this exists as ContractContext
+      --        however that lacks a field for the BlockHashInput.
+      --        which is why this input type is here (for now).
```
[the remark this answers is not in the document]
vikt0r0: It's not a problem? :-)

src/Concordium/Client/GRPC2.hs
```diff
@@ -0,0 +1,2693 @@
...
+        case r of
+            Proto.RejectReason'ModuleNotWf _ ->
+                return ModuleNotWF
+            Proto.RejectReason'ModuleHashAlreadyExists mr ->
+                ModuleHashAlreadyExists <$> fromProto mr
+            Proto.RejectReason'InvalidAccountReference addr ->
+                InvalidAccountReference <$> fromProto addr
+            Proto.RejectReason'InvalidInitMethod' rr -> do
+                moduleRef <- fromProto =<< rr ^? ProtoFields.moduleRef
+                initName <- fromProto =<< rr ^? ProtoFields.initName
```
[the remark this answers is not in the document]
vikt0r0: I thought about this as well. If all fields are guaranteed to be present in the message, sure, I will opt for doing this all over.

src/Concordium/Client/GRPC2.hs
```diff
@@ -0,0 +1,2693 @@
...
+        signatures <-
+            fmap Map.fromAscList
+                . mapM deMk
+                . Map.toAscList
+                =<< uis ^? ProtoFields.signatures
+        return Updates.UpdateInstructionSignatures{..}
+      where
+        deMk (k, s) = do
+            sig <- fromProto s
+            return (fromIntegral k, sig)
```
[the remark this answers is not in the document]
vikt0r0: It definitely definitely is not. `k` is a `Word32` and the output is `Word16`, so a check will be added.

src/Concordium/Client/GRPC2.hs
```diff
@@ -0,0 +1,2693 @@
...
+
+instance FromProto Proto.NewRelease where
+    type Output' Proto.NewRelease = (Timestamp, Amount)
+    fromProto nRelease = do
+        tStamp <- fmap fst . fromProto =<< nRelease ^? ProtoFields.timestamp
+        amount <- fromProto =<< nRelease ^? ProtoFields.amount
+        return (tStamp, amount)
+
+instance FromProto Proto.AccountTransactionDetails where
+    type Output' Proto.AccountTransactionDetails = (Maybe TransactionType, ValidResult)
```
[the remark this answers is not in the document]
vikt0r0: There is a cost field in the parent (`BlockItemSummary'AccountTransaction`) message as well (This is the message which is translated to `TransactionSummary`). Do you mean that I should propagate this value to the caller and set it there?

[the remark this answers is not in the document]
abizjak: Do you mean the `energy_cost`? All block items have that, but only account transactions have the cost, which is an amount that was charged to the sender.

What I meant was that either you also return the cost here, so the return type here is a triple (Maybe ..., ValidResult, Amount).

Or you don't have this conversion as an instance, and instead just have a helper function that you use when translating BlockItem.

src/Concordium/Client/GRPC2.hs
```diff
@@ -0,0 +1,2693 @@
...
+
+instance FromProto Proto.UpdateInstruction where
+    type Output' Proto.UpdateInstruction = Updates.UpdateInstruction
+    fromProto ui = do
+        uiSignHash <- Nothing -- FIXME: What should go here, no such field in the protobuf equivalent.
+        uiSignatures <- fromProto =<< ui ^? ProtoFields.signatures
+        uiHeader <- fromProto =<< ui ^? ProtoFields.header
+        uiPayload <- do
+            pl <- ui ^? (ProtoFields.payload . ProtoFields.rawPayload)
+            case (S.runGet $ Updates.getUpdatePayload SP1) pl of -- FIXME: What protocol version to use here?
```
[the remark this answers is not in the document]
vikt0r0: So if the payload is decoded w.r.t a specific protocol version, it sounds like we should add a field for either the PV or the undecoded payload in protocol buffer schema and include that in the message from the node side of things?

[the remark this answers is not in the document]
abizjak: This is just an artifact of how it is modelled in Haskell.

The simplest solution is to add a version of UpdateInstruction that has undecoded payload.

The only other option would be that you know the protocol version here. That you could technically do but it will be messy and involve boilerplate code.

Since this endpoint does not exist for V1, and is thus currently not used by anything in concordium-client, please put it on hold. We will handle it in some followup work. I don't want to delay migration of the client any further because of this.

src/Concordium/Client/GRPC2.hs
```diff
@@ -0,0 +1,2766 @@
...
+instance ToProto BlockHashInput where
+    type Output BlockHashInput = Proto.BlockHashInput
+    toProto = \case
+        Best -> Proto.make $ ProtoFields.best .= defMessage
+        LastFinal -> Proto.make $ ProtoFields.lastFinal .= defMessage
+        Given bh -> Proto.make $ ProtoFields.given .= toProto bh
+
+instance FromProto Proto.BlockHash where
+    type Output' Proto.BlockHash = BlockHash
+    fromProto bh = BlockHash <$> deMkSerialize bh
```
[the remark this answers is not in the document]
vikt0r0: Yes, as mentioned `S.decode` fails. I tested it with `getBlocks`.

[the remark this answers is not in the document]
abizjak: What is the raw value you are converting?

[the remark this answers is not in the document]
vikt0r0: Here are some examples of the received blockhash bytestrings dumped as hex/escaped octal(?):

```
Hex: \"3199f6f27393783c7fb970054b39a0d9b3177c3e605a66931b8fbbee9141db5f\"
Raw: 1\153\246\242s\147x<\DEL\185p\ENQK9\160\217\179\ETB|>`Zf\147\ESC\143\187\238\145A\219_

Hex: \"f95cb13af02e9a1275fa6a0a58833a8eb304d0784097e7e17eab9d10cd1901ea\"
Raw: \249\\\177:\240.\154\DC2u\250j\nX\131:\142\179\EOT\208x@\151\231\225~\171\157\DLE\205\EM\SOH\234

Hex: \"64b09e4d666600907ab37756428e58664122e5bddd0d738f12acf1ee20a024bf\"
Raw: d\176\158Mff\NUL\144z\179wVB\142XfA\"\229\189\221\rs\143\DC2\172\241\238 \160$\191

Hex: \"1cd7b1debe02715a09a629b338643b1bdd63c649570771e8289ba7391388ecd8\"
Raw: \FS\215\177\222\190\STXqZ\t\166)\179\&8d;\ESC\221c\198IW\aq\232(\155\167\&9\DC3\136\236\216

Hex: \"813b728c41921b207b79cfc5bdea2db9a933d846228ec0671f4ce447906b71a4\"
Raw: \129;r\140A\146\ESC {y\207\197\189\234-\185\169\&3\216F\"\142\192g\USL\228G\144kq\164
```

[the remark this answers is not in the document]
abizjak: Are you sure that it is this conversion that does not work?

Because
```
$ Data.Serialize.decode (fromString "d\176\158Mff\NUL\144z\179wVB\142XfA\"\229\189\221\rs\143\DC2\172\241\238 \160$\191") :: Either String Concordium.Crypto.SHA256.Hash
Right 64b09e4d666600907ab37756428e58664122e5bddd0d738f12acf1ee20a024bf
```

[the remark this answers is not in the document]
vikt0r0: Of course this is due to a bug I have in`deMkSerialize`, it does not wrap the return value in the second case:

```
deMkSerialize val = case S.decode $ val ^. ProtoFields.value of
    Left _ -> Nothing
    Right v -> v
```

Sorry about the the confusion.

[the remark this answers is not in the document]
abizjak: Another instance of "code that sufficently generic in types is very similar to untyped code" :)

src/Concordium/Client/GRPC2.hs
```diff
@@ -0,0 +1,2766 @@
...
+            ProtoFields.amount .= toProto ccAmount
+            ProtoFields.entrypoint .= toProto ccMethod
+            ProtoFields.parameter .= toProto ccParameter
+            ProtoFields.energy .= toProto ccEnergy
+
+instance FromProto Proto.ContractEvent where
+    type Output' Proto.ContractEvent = Wasm.ContractEvent
+    -- FIXME: Can we just use deMkSerialize here? Types check out, but fromShort
+    -- is used in the other direction...
+    fromProto ce = return . Wasm.ContractEvent . BSS.toShort $ ce ^. ProtoFields.value
```
abizjak, line 911: No, we cannot use serialization here. Serialization puts a length header first, and thus also expects it when parsing.

src/Concordium/Client/GRPC2.hs
```diff
@@ -0,0 +1,2766 @@
...
+
+instance FromProto Proto.Parameter where
+    type Output' Proto.Parameter = Wasm.Parameter
+    -- FIXME: There are protocol-dependent limits on the length. Can something
+    -- be done to enforce this here?
```
abizjak, line 916: No. Since everything is at protocol 5 now the limit is the more generous one.

src/Concordium/Client/GRPC2.hs
```diff
@@ -0,0 +1,2766 @@
...
+    -- FIXME: Can we just use deMkSerialize here? Types check out, but fromShort
+    -- is used in the other direction...
```
abizjak, line 918: No, we cannot for the same reason as for contract events.

src/Concordium/Client/GRPC2.hs
```diff
@@ -0,0 +1,2766 @@
...
+    toProto ip = Proto.make $ ProtoFields.value .= ip
+
+instance ToProto IpPort where
+    type Output IpPort = Proto.Port
+    toProto ip = Proto.make $ ProtoFields.value .= fromIntegral ip
+
+instance FromProto Proto.IpAddress where
+    type Output' Proto.IpAddress = IpAddress
+
+    -- FIXME: Should this be validated?
```
abizjak, line 1202: Not now. this is just going to be printed.

If we did we should get find a library (does prelude define it, or network?) that defines an IP data type. We are definitely not writing our own IP address validation.

abizjak: Anyway, just ignore validation here since you are mapping it to a string anyhow.

src/Concordium/Client/GRPC2.hs
```diff
@@ -0,0 +1,2766 @@
...
+        _nusnAddIdentityProvider <- fromProto =<< nums ^? ProtoFields.addIdentityProvider
+        _nusnCooldownParameters <- fromProto =<< nums ^? ProtoFields.cooldownParameters
+        _nusnTimeParameters <- fromProto =<< nums ^? ProtoFields.timeParameters
+        return QueryTypes.NextUpdateSequenceNumbers{..}
+
+-- An IP address.
+type IpAddress = Text
+
+-- An IP port.
+type IpPort = Int16
```
abizjak, line 1183: Ports should be Word16.

src/Concordium/Client/GRPC2.hs
```diff
@@ -0,0 +1,2766 @@
...
+    type Output' Proto.IpAddress = IpAddress
+
+    -- FIXME: Should this be validated?
+    fromProto peer = peer ^? ProtoFields.value
+
+instance FromProto Proto.Port where
+    type Output' Proto.Port = IpPort
+
+    -- FIXME: Should this be validated?
+    fromProto port = fromIntegral <$> port ^? ProtoFields.value
```
abizjak, line 1209: Well just make sure it fits as is already done for many other values.

src/Concordium/Client/GRPC2.hs
```diff
@@ -0,0 +1,2766 @@
...
+    type Output' Proto.BannedPeer = Peer
+    fromProto peer = fromProto =<< peer ^? ProtoFields.ipAddress
+
+instance FromProto Proto.BannedPeers where
+    type Output' Proto.BannedPeers = [Peer]
+    fromProto peers = mapM fromProto =<< peers ^? ProtoFields.peers
+
+instance FromProto Proto.InstanceStateValueAtKey where
+    type Output' Proto.InstanceStateValueAtKey = ByteString
+    fromProto = deMkSerialize
```
abizjak, line 1221: I don't think this is correct. The value is the bytestring. It does not need to be deserialized. And if you try it will fail since bytestring serialization uses 8 bytes for length and then the actual string.

src/Concordium/Client/GRPC2.hs
```diff
@@ -0,0 +1,2766 @@
...
+
+instance FromProto Proto.Signature where
+    type Output' Proto.Signature = Signature
+    fromProto = deMkSerialize
```
abizjak, line 1225: This is not correct. The ToProto just does fromShort.

```
instance ToProto Signature where
    type Output Signature = Proto.Signature

    toProto (Signature bss) = Proto.make $ do
        ProtoFields.value .= BSS.fromShort bss

```

src/Concordium/Client/GRPC2.hs
```diff
@@ -0,0 +1,2766 @@
...
+        bakerId <- fromProto =<< bkEvent ^? ProtoFields.bakerId
+        addr <- fromProto =<< bkEvent ^? ProtoFields.account
+        signKey <- fromProto =<< bkEvent ^? ProtoFields.signKey
+        electionKey <- fromProto =<< bkEvent ^? ProtoFields.electionKey
+        aggregationKey <- fromProto =<< bkEvent ^? ProtoFields.aggregationKey
+        return (bakerId, addr, signKey, electionKey, aggregationKey)
+
+instance FromProto Proto.RegisteredData where
+    type Output' Proto.RegisteredData = RegisteredData
+    -- FIXME: Can we just use deMkSerialize here? Types check out, but fromShort
```
abizjak, line 1332: No, we cannot.

src/Concordium/Client/GRPC2.hs
```diff
@@ -0,0 +1,2766 @@
...
+    type Output' Proto.NewRelease = (Timestamp, Amount)
+    fromProto nRelease = do
+        tStamp <- fromProto =<< nRelease ^? ProtoFields.timestamp
+        amount <- fromProto =<< nRelease ^? ProtoFields.amount
+        return (tStamp, amount)
+
+instance FromProto Proto.Memo where
+    type Output' Proto.Memo = Memo
+    -- FIXME: Can we just use deMkSerialize here? Types check out, but fromShort
+    -- is used in the other direction...
```
abizjak, line 1346: No.

src/Concordium/Client/GRPC2.hs
```diff
@@ -0,0 +1,2766 @@
...
+    fromProto memo = Memo . BSS.toShort <$> memo ^? ProtoFields.value
+
+instance FromProto Proto.ArInfo'ArIdentity where
+    type Output' Proto.ArInfo'ArIdentity = ArIdentity
+    fromProto arIdentity = do
+        let arId = deMkWord32 arIdentity
+        -- This can not be 0.
+        -- FIXME: This is not checked in the FFI `ar_info_create` function of base.
```
abizjak, line 1354: There is nothing to fix. It is checked whenever external data is used to create this, in deserialization, and now here. We don't need to check it everywhere internally.

src/Concordium/Client/GRPC2.hs
```diff
@@ -0,0 +1,2766 @@
...
+            else return $ ArIdentity arId
+
+instance FromProto Proto.IpIdentity where
+    type Output' Proto.IpIdentity = IdentityProviderIdentity
+    fromProto arIdentity = do
+        let ipIp = deMkWord32 arIdentity
+        -- FIXME: Can this not be 0 either?
+        -- FIXME: This is not checked in the FFI `ip_info_create` function of base.
+        if ipIp == 0
+            then Nothing
```
abizjak, line 1366: This can be 0, and is 0 on mainnet and testnet for an identity provider.

There are technical reasons for this being allowed, and not allowed for ArIdentity.

src/Concordium/Client/GRPC2.hs
```diff
@@ -0,0 +1,2766 @@
...
+
+instance FromProto Proto.TransactionFeeDistribution where
+    type Output' Proto.TransactionFeeDistribution = Parameters.TransactionFeeDistribution
+    fromProto tfDistribution = do
+        _tfdBaker <- fromProto =<< tfDistribution ^? ProtoFields.baker
+        _tfdGASAccount <- fromProto =<< tfDistribution ^? ProtoFields.gasAccount
+        return Parameters.TransactionFeeDistribution{..}
+
+instance FromProto Proto.UpdatePayload where
+    type Output' Proto.UpdatePayload = (Updates.UpdateType, Updates.UpdatePayload)
```
abizjak, line 1642: There is a function
```
updateType :: UpdatePayload -> UpdateType
```
in base.

Please use that and do not reinvent this here since that is error prone. Just return UpdatePayload.

src/Concordium/Client/GRPC2.hs
```diff
@@ -0,0 +1,2766 @@
...
+        tsIndex <- TransactionIndex . deMkWord64 <$> biSummary ^? ProtoFields.index
+        tsEnergyCost <- fromProto =<< biSummary ^? ProtoFields.energyCost
+        tsHash <- fromProto =<< biSummary ^? ProtoFields.hash
+        -- Discern between transactions
+        bis <- biSummary ^. Proto.maybe'details
+        case bis of
+            -- Account creation
+            ProtoFields.BlockItemSummary'AccountCreation aCreation -> do
+                let tsSender = Nothing
+                tsCost <- Just $ Amount 0
```
abizjak, line 1724:
```suggestion
                let tsCost = 0
```

src/Concordium/Client/GRPC2.hs
```diff
@@ -0,0 +1,2766 @@
...
+            ProtoFields.BlockItemSummary'AccountTransaction aTransaction -> do
+                sender <- aTransaction ^? ProtoFields.sender
+                let tsSender = fromProto sender
+                tsCost <- fromProto =<< aTransaction ^? ProtoFields.cost
+                (tType, tsResult) <- fromProto (sender, aTransaction)
+                let tsType = TSTAccountTransaction tType
+                return TransactionSummary{..}
+            ProtoFields.BlockItemSummary'Update update -> do
+                let tsSender = Nothing
+                tsCost <- Just $ Amount 0
```
abizjak, line 1739:
```suggestion
                tsCost <- Just $ Amount 0
```

src/Concordium/Client/GRPC2.hs
```diff
@@ -0,0 +1,2766 @@
...
+                ebsreRestakeEarnings <- breUpdated ^? ProtoFields.restakeEarnings
+                return BakerSetRestakeEarnings{..}
+            ProtoFields.BakerEvent'BakerKeysUpdated bkUpdated -> do
+                (ebkuBakerId, ebkuAccount, ebkuSignKey, ebkuElectionKey, ebkuAggregationKey) <- fromProto bkUpdated
+                return BakerKeysUpdated{..}
+            ProtoFields.BakerEvent'BakerSetOpenStatus' bsoStatus -> do
+                let ebsosAccount = sender
+                ebsosBakerId <- fromProto =<< bsoStatus ^? ProtoFields.bakerId
+                ebsosOpenStatus <- fromProto =<< bsoStatus ^? ProtoFields.openStatus
+                return BakerSetOpenStatus{..}
```
abizjak, line 1957: Same issue with `^?` on a Lens.

src/Concordium/Client/GRPC2.hs
```diff
@@ -0,0 +1,2766 @@
...
+      consensusInfo :: !(Maybe PeerCatchupStatus)
+    }
+
+-- |Network related peer statistics.
+type PeersInfo = [PeerInfo]
+
+instance FromProto Proto.IpSocketAddress where
+    type Output' Proto.IpSocketAddress = IpSocketAddress
+    fromProto ipsAddress = do
+        -- FIXME: Should this be validated?
```
abizjak, line 2081: Not at this place

src/Concordium/Client/GRPC2.hs
```diff
@@ -0,0 +1,2766 @@
...
+                    return ActiveBakerCommitteeInfo
+                ProtoFields.NodeInfo'BakerConsensusInfo'ActiveFinalizerCommitteeInfo' _ ->
+                    return ActiveFinalizerCommitteeInfo
+                ProtoFields.NodeInfo'BakerConsensusInfo'PassiveCommitteeInfo pcInfo ->
+                    PassiveBaker <$> fromProto pcInfo
+        return BakerConsensusInfo{..}
+
+-- |Consensus info for a node configured with baker keys.
+data BakerConsensusInfo = BakerConsensusInfo
+    { bakerId :: !Integer,
```
abizjak, line 2175: This is, as the name suggests, a `BakeId`, not an `Integer`.

src/Concordium/Client/GRPC2.hs
```diff
@@ -0,0 +1,2766 @@
...
+      -- has responded to the request, its status will be changed to either `UpToDate`
+      -- or `Pending`.
+      CatchingUp
+
+-- |Network statistics for the peer.
+data NetworkStats = NetworkStats
+    { -- |The number of messages sent to the peer.
+      -- Packets are blocks, transactions, catchup messages, finalization records
+      -- and network messages such as pings and peer requests.
+      packetsSent :: !Integer,
```
abizjak, line 2058: All of these are Word64.

src/Concordium/Client/GRPC2.hs
```diff
@@ -0,0 +1,2766 @@
...
+            return $ Parameters.CooldownParametersV0 epoch
+        let _cpTimeParameters = Parameters.TimeParametersV0
+        _cpAccountCreationLimit <- fromProto =<< cParams ^? ProtoFields.accountCreationLimit
+        _cpRewardParameters <- do
+            _rpMintDistribution <- fromProto =<< cParams ^? ProtoFields.mintDistribution
+            _rpTransactionFeeDistribution <- fromProto =<< cParams ^? ProtoFields.transactionFeeDistribution
+            _rpGASRewards <- fromProto =<< cParams ^? ProtoFields.gasRewards
+            return Parameters.RewardParameters{..}
+        -- FIXME: ProtoFields.foundationAccount has type AccountAddress, but the type expects an AccountIndex.
+        --        Is it possible to convert this?
```
abizjak, line 2274: Not without context. You can look up an address via an index using `GetAccountInfo`. But you cannot do this without a network call since it depends on node state.

You should probably return an account index here, and a closure that takes an account address and returns chain parameters.

Then you can query account info together with chain parameters to get all the relevant data.

src/Concordium/Client/GRPC2.hs
```diff
@@ -0,0 +1,2766 @@
...
+        _cpTimeParameters <- fromProto =<< cParams ^? ProtoFields.timeParameters
+        _cpAccountCreationLimit <- fromProto =<< cParams ^? ProtoFields.accountCreationLimit
+        _cpRewardParameters <- do
+            _rpMintDistribution <- fromProto =<< cParams ^? ProtoFields.mintDistribution
+            _rpTransactionFeeDistribution <- fromProto =<< cParams ^? ProtoFields.transactionFeeDistribution
+            _rpGASRewards <- fromProto =<< cParams ^? ProtoFields.gasRewards
+            return Parameters.RewardParameters{..}
+        -- FIXME: ProtoFields.foundationAccount has type AccountAddress, but the type expects an AccountIndex.
+        --        Is it possible to convert this?
+        _cpFoundationAccount <- Nothing -- <-|
```
abizjak, line 2304: Same here.

src/Concordium/Client/GRPC2.hs
```diff
@@ -0,0 +1,2766 @@
...
+-- only be sent by the governance committee.
+--
+-- Returns a hash of the block item, which can be used with
+-- `GetBlockItemStatus`.
+sendBlockItemV2 :: (MonadIO m) => SendBlockItemInput -> ClientMonad m (GRPCResult (Maybe TransactionHash))
+sendBlockItemV2 sbiInput = withUnaryCoreV2 (callV2 @"sendBlockItem") msg ((fmap . fmap) fromProto)
+  where
+    msg = toProto sbiInput
+
+-- FIXME: Ergonomics: Should input, resp. output be ByteString?'
```
abizjak, line 2414: Yes, they should be. Keys and values are bytestrings.

src/Concordium/Client/GRPC2.hs
```diff
@@ -0,0 +1,2766 @@
...
+withUnaryV2 ::
+    forall m n b.
+    (HasMethod CS.Queries m, MonadIO n) =>
+    RPC CS.Queries m ->
+    MethodInput CS.Queries m ->
+    SimpleGetter (GRPCResponse (MethodOutput CS.Queries m)) b ->
+    ClientMonad n (Either String b)
+withUnaryV2 method message k = withUnaryCoreV2 method message (\x -> (^. k) <$> x)
+
+-- |Call a method with a given message using the `id` lens on the output.
```
abizjak, line 2726:
```suggestion
-- |Call a method with a given message without processing the output.
```

abizjak (changes requested):
There are still a few critical issues remaining.

And there is this general issue that you use `^?` a lot on a lens, but I don't see how that is necessary. It is a Lens, so you should be able to use `^.`.

vikt0r0:
> There are still a few critical issues remaining.
> 
> And there is this general issue that you use ^? a lot on a lens, but I don't see how that is necessary. It is a Lens, so you should be able to use ^..

Great. I annotated the places where I had doubts. Thanks for the feedback 👍🏼 🙂

The why: The `proto-lens` docs kind of suck, and I opted for `^?` as it was listed as safe. When I started this, it was unclear whether I could expect all message fields to be present. Search-replace with a well-crafted regex should remedy this, and should not impact things too much.

src/Concordium/Client/GRPC2.hs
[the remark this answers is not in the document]
vikt0r0: Right. As mentioned I think `Output` is a fine choice for naming this.

src/Concordium/Client/GRPC2.hs
vikt0r0: So the limit to be used is the V5 one. And it will only increase monotonically over time?

src/Concordium/Client/GRPC2.hs
vikt0r0: I see. Since this is a non-optional field, should it be set to a dummy value then?

src/Concordium/Client/GRPC2.hs
Bargsteen: I don't think we want to make any promises, but I assume that will be the case, yes ;)

src/Concordium/Client/GRPC2.hs
```diff
@@ -0,0 +1,2766 @@
...
+--
+-- More concretely, the wrapper type should be of the form
+--
+-- > message Wrapper {
+-- >    bytes value = 1
+-- > }
+--
+-- where the name @Wrapper@ can be arbitrary, but the @value@ field must exist,
+-- and it must have type @bytes@.
+deMkSerialize ::
```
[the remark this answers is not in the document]
vikt0r0: I believe it would be a bit misleading to prefix it with `mk`. Unless I am misunderstood, the `mk` prefix is of `Proto.make`-fame and refers to constructing a message with the `value` field, and `Serialize` refers to `S.encode`, hence `mkSerialize`.

The `de`-prefix is merely to indicate it as an inverse. If you still  think `mkDeserialize` is better, I will change it to that 🙂

src/Concordium/Client/GRPC2.hs
vikt0r0: Nice. Thanks.

src/Concordium/Client/GRPC2.hs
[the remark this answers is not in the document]
abizjak: It's fine, although what you describe is not really the etymology of the name. `mk` is not to do with `Proto.make`.

I don't want to discuss naming further.

src/Concordium/Client/GRPC2.hs
abizjak: Sounds reasonable.

src/Concordium/Client/GRPC2.hs
```diff
@@ -0,0 +1,2766 @@
...
+        pCreatedAt <- fromProto =<< p ^? ProtoFields.createdAt
+        pValidTo <- fromProto =<< p ^? ProtoFields.validTo
+        pItems <-
+            fmap Map.fromAscList
+                . mapM convert
+                . Map.toAscList
+                =<< p ^? ProtoFields.attributes
+        return Policy{..}
+      where
+        -- \|Convert a tag and a bytestring pair to an attribute tag and attribute value pair.
```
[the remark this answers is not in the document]
vikt0r0: This kept popping up due to fourmolu acting up for reasons I do not understand. I have removed it now.

src/Concordium/Client/GRPC2.hs
```diff
@@ -0,0 +1,2766 @@
...
+    ( Data.ProtoLens.Field.HasField
+        b
+        "value"
+        Word32
+    ) =>
+    b ->
+    Word32
+deMkWord32 val = coerce $ val ^. ProtoFields.value
+
+-- |Like 'deMkWord64' above, but for Word16 fields instead.
```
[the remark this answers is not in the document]
abizjak: Right, but you having to explain a comment in another comment is surely a sign that the original comment is misleading?

This function is really not "like" deMkWord64. This one additionally checks ranges.

Admittedly this is a very minor issue, but please try to think of the reader when writing comments.

src/Concordium/Client/GRPC2.hs
```diff
@@ -0,0 +1,2658 @@
...
+
+instance FromProto Proto.AccountAddress where
+    type Output Proto.AccountAddress = AccountAddress
+    fromProto = deMkSerialize
+
+instance FromProto Proto.CredentialRegistrationId where
+    type Output Proto.CredentialRegistrationId = (CredentialRegistrationID, RawCredentialRegistrationID)
+    fromProto crId = do
+        raw <- deMkSerialize crId
+        nonRaw <- credIdFromRaw raw
```
abizjak, line 213: Where is this function defined? There is no PR in base, nor is this in main there.

src/Concordium/Client/GRPC2.hs
```diff
@@ -0,0 +1,2658 @@
...
+        signatures <-
+            fmap Map.fromAscList
+                . mapM deMk
+                . Map.toAscList
+                $ uis ^. ProtoFields.signatures
+        return Updates.UpdateInstructionSignatures{..}
+      where
+        deMk (k, s) = do
+            -- Ensure that the value fits into a Word16.
+            if fromIntegral k <= (maxBound :: Word16)
```
abizjak, line 1135: This check will always succeed since you are first converting a value to Word16, and then checking if the value is less than maxBound :: Word16. Which is always true.

You need to
```
k <= fromIntegral (maxBound :: Word16)
```

instead.

src/Concordium/Client/GRPC2.hs
```diff
@@ -0,0 +1,2658 @@
...
+
+instance FromProto Proto.ContractStateV0 where
+    type Output Proto.ContractStateV0 = Wasm.ContractState
+    fromProto cs = return . Wasm.ContractState $ cs ^. ProtoFields.value
+
+instance FromProto Proto.ReceiveName where
+    type Output Proto.ReceiveName = Wasm.ReceiveName
+    fromProto name = do
+        let n = name ^. ProtoFields.value
+        unless (Wasm.isValidReceiveName n) Nothing
```
abizjak, line 593:
```suggestion
        guard (Wasm.isValidReceiveName n)
```

src/Concordium/Client/GRPC2.hs
```diff
@@ -0,0 +1,2658 @@
...
+        return $ Wasm.ReceiveName n
+
+instance FromProto Proto.InitName where
+    type Output Proto.InitName = Wasm.InitName
+    fromProto name = do
+        let n = name ^. ProtoFields.value
+        unless (Wasm.isValidInitName n) Nothing
```
abizjak, line 600:
```suggestion
        guard (Wasm.isValidInitName n)
```

abizjak (approved):

src/Concordium/Client/GRPC2.hs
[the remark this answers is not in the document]
vikt0r0: I see. I'll mention that in the comment.

vikt0r0:
In a feature branch, you can click the folder in GitHub to see the commit ref:

https://github.com/Concordium/concordium-base/tree/d132f5de85753b3ac7200c60ba5f756eef270dd3

So we should also have have PR’s for things that are not PRs pulling into main?

src/Concordium/Client/GRPC2.hs
vikt0r0: Right! Nice catch.

src/Concordium/Client/GRPC2.hs
vikt0r0: It is in a feature branch, you can click the folder in GitHub to see the commit ref, or it is here:

https://github.com/Concordium/concordium-base/tree/d132f5de85753b3ac7200c60ba5f756eef270dd3

So I guess we should also have have PR’s for things that are not PRs that pull into main then, right?

vikt0r0: Here, specifically: https://github.com/Concordium/concordium-base/blob/d132f5de85753b3ac7200c60ba5f756eef270dd3/haskell-src/Concordium/ID/Types.hs#L952

abizjak: Right, but that is not merged to main, right? Nor is there a PR? Or where is this?

vikt0r0:
No, nothing is merged into main.

PR #210, opened 2023-02-02
Implement streaming V2 GRPC API calls.

vikt0r0:
### Purpose

This PR concerns itself with implementing the _streaming_ V2 GRPC node API calls. The changes implement helpers for interacting with the streaming API calls of the V2 GRPC API as well as functionality for translating between generated protocol buffer schema types and their native (Concordium) Haskell type equivalents where relevant. Specifically, this translation is used for interface parameters and return values.

### Depends on
https://github.com/Concordium/concordium-base/pull/209

### Changes
Implementation of helpers and API calls: 
  - **Implement helper function for handling streaming GRPC endpoints**
  - **Implement (streaming) V2 GRPC API calls**
    - getBlocks
    - getFinalizedBlocks
    - getAccountList
    - getModuleList
    - getAncestors
    - getInstanceList
    - getInstanceState (unused in client commands)
    - getBakerList
    - getPoolDelegators
    - getPoolDelegatorsRewardPeriod
    - getPassiveDelegators
    - getPassiveDelegatorsRewardPeriod
    - getIdentityProviders
    - getAnonymityRevokers
    - getAccountNonFinalizedTransactions
    - getBlockTransactionEvents
    - getBlockSpecialEvents
    - getBlockPendingUpdates


vikt0r0:
@Bargsteen I hope it's alright I requested your review for this PR, as it is a continuation of (and quite similar to) [this](https://github.com/Concordium/concordium-client/pull/209). And otherwise just let me know. In any case, thanks 🙂

src/Concordium/Client/GRPC2.hs
```diff
@@ -2552,20 +2852,234 @@ getAccountInfoV2 ::
...
+    RPC CS.Queries m ->
+    -- |The procedure input.
+    i ->
+    -- |A mapping of the result.
+    (GRPCResult o -> b) ->
+    ClientMonad n b
+withUnaryV2 method input k = withGRPCCoreV2 callHelper k
+    where
+        -- This is here so we may leverage `withGRPCCoreV2`.
+        callHelper client = do
```
abizjak, line 2885: Yeah, we do need to handle failures and such like we do in the V1 version.

src/Concordium/Client/GRPC2.hs
```diff
@@ -2552,20 +2852,234 @@ getAccountInfoV2 ::
...
+    ClientMonad n b
+withServerStreamCallbackV2 method input acc handler k =
+    withGRPCCoreV2 callHelper k
+    where
+        -- This is simply a handler which conforms to the one required by
+        -- `rawStreamServer`. This is here, so we may ignore the response
+        -- headers in the handler, just for the sake of ergonomics.
+        handler' acc' _hds streamObj = handler acc' streamObj
+        -- Helper that invokes the streaming call with the above callback.
+        -- This is here so we may leverage `withGRPCCoreV2`.
```
abizjak, line 2937: The thing is, withGRPCCoreV2 is a bit tricky to use for streaming calls.

It is fine for calls that will immediately collect everything. But not for things like getBlocks and getFinalizedBlocks.

The reason being that it uses a RW lock internally. Meaning that during that stream you will not be able to handle any failures and reconnect.

In the V1 it made sense since queries were all unary. And it also makes sense I suppose for streaming where you collect immediately.

src/Concordium/Client/GRPC2.hs
```diff
@@ -2552,20 +2852,234 @@ getAccountInfoV2 ::
...
+        callHelper client = do
+            res <- rawStreamServer method client acc input handler'
+            return $ fmap ServerStreamOutput res
+
+-- |Setup a GRPC client and invoke a request helper which uses the client.
+-- The output is interpreted using the function given in the second parameter.
+withGRPCCoreV2 ::
```
abizjak, line 2944: I am not a fan of copying this. THis function is subtle. Can we not just reuse the one we already have for V1?

src/Concordium/Client/GRPC2.hs
```diff
@@ -2359,42 +2360,341 @@ instance ToProto SendBlockItemInput where
...
+
+-- |Get all smart contract modules in a given block.
+getModuleListV2 :: (MonadIO m) => BlockHashInput -> ClientMonad m (GRPCResult (Maybe (Seq ModuleRef)))
+getModuleListV2 blockHash = withServerStreamCollectV2 (callV2 @"getModuleList") msg ((fmap . fmap . mapM) fromProto)
+  where
+    msg = toProto blockHash
+
+-- |Get all accounts that exist at the end of a given block.
+getAccountListV2 :: (MonadIO m) => BlockHashInput -> ClientMonad m (GRPCResult (Maybe (Seq AccountAddress)))
+getAccountListV2 blockHash = withServerStreamCollectV2 (callV2 @"getAccountList") msg ((fmap . fmap . mapM) fromProto)
```
abizjak, line 2646: None of these arguments are block hashes. So please name them appropriately.

src/Concordium/Client/GRPC2.hs
```diff
@@ -2359,42 +2360,341 @@ instance ToProto SendBlockItemInput where
...
+
+-- |Get all identity providers registered at the end of a given block.
+getIdentityProvidersV2 :: (MonadIO m) => BlockHashInput -> ClientMonad m (GRPCResult (Maybe (Seq IpInfo.IpInfo)))
+getIdentityProvidersV2 blockHash =
+    withServerStreamCollectV2 (callV2 @"getIdentityProviders") msg ((fmap . fmap . mapM) fromProto)
+  where
+    msg = toProto blockHash
+
+-- |Get all fixed passive delegators for the reward period of a given block.
+-- In contrast to `getPssiveDelegatorsV2` which returns all delegators registered
```
abizjak, line 2571:
```suggestion
-- In contrast to 'getPassiveDelegatorsV2' which returns all delegators registered
```

src/Concordium/Client/GRPC2.hs
vikt0r0: They are identical up to a few changes. Specifically it uses `outputGRPCV2` rather than `outputGRPC`, and a helper rather than `rawUnary`. It's technicallt copied sincr I commented out the old version in the commit. It is not possible to use the V1 version for the streaming calls though. What do you propose?

abizjak: So the way the client is used now it is good enough I think. The only time we would use "getBlocks/getFinalizedBlocks" is for tailing transactions, which means nothing else is happening then anyhow.

Also the way it is used in the wallet-proxy it is fine. So we will probably have to just live with this in order to get this done. It is unfortunate the Haskell library is shitty. I don't want to spend much time revising this further unless it becomes an issue.

So for start we make it clear that there is this limitation and move on.

abizjak: Did this answer your question?

vikt0r0: I believe so 👍🏼 so do I understand this correctly: The issue comes from the client instance currently working like a singleton with a mutex?

So this is okay for the _non-streaming_ and _non-tailing, streaming_ calls as their duration for acquiring the mutex is likely to be minute.

For tailing calls, the mutex is acquired for the duration of the call, but I suppose this is OK since these are not used in the wallet-proxy.

So this limitation should be stated in the documentation (perhaps) of `withGRPCCoreV2` and perhaps `getBlocksV2`/`getFinalizedBlocksV2`, for future reference.

vikt0r0: Do we need to set a request timeout for the tailing calls?

abizjak: It's not a mutex in normal operation. It is reader-writer lock protected.  So multiple queries are supported in paralell, but if things go wrong then we need to re-establish the connection and that requires a writer lock.

The issue is that if there is a long-running (and non-productive) streaming call then it will have the reader lock acquired for the full duration of the stream. Which will prevent connection-reset since the writer lock cannot be acquired.

It should not be an issue in practice the way the client is used, we just need to be aware of it.

abizjak: > Do we need to handle request timeout differently for the tailing calls?

Yeah.

But let's not go down that rabit hole now. Let's do this as a last thing after we have a working client. We don't need the getFinalizedBlocks and getBlocks calls to get a working client mirroring what it does now. We can continue polling as we do now for transaction statuses.

vikt0r0: > But let's not go down that rabit hole now.

Right, just as to make note of this.

src/Concordium/Client/GRPC2.hs
```diff
@@ -2359,42 +2360,341 @@ instance ToProto SendBlockItemInput where
...
+        return ArrivedBlockInfo{..}
+
+instance FromProto Proto.InstanceStateKVPair where
+    -- FIXME: Should this be ByteStrings, or something else? I know this
+    --        isn't needed, but I added it before realizing. Let me know
+    --        if it should just be deleted.
+    type Output' Proto.InstanceStateKVPair = (ByteString, ByteString)
+    fromProto ikvPair = do
+        key <- ikvPair ^? ProtoFields.key
+        value <- ikvPair ^? ProtoFields.value
```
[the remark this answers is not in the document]
vikt0r0: It definitely is not! I will do a pass where I replace them once everything else looks OK.

src/Concordium/Client/GRPC2.hs
```diff
@@ -2359,42 +2360,343 @@ instance ToProto SendBlockItemInput where
...
+
+-- |Get IDs of all bakers at the end of a given block.
+getBakerListV2 :: (MonadIO m) => BlockHashInput -> ClientMonad m (GRPCResult (Maybe (Seq.Seq BakerId)))
+getBakerListV2 bhInput = withServerStreamCollectV2 (callV2 @"getBakerList") msg ((fmap . fmap . mapM) fromProto)
+  where
+    msg = toProto bhInput
+
+-- |Get key-value pairs representing the entire state of a specific contract instance in a given block.
+-- The resulting sequence is ordered lexicographically according to its keys.
+getInstanceStateV2 :: (MonadIO m) => BlockHashInput -> ContractAddress -> ClientMonad m (GRPCResult (Maybe (Seq.Seq (ByteString, ByteString))))
```
Bargsteen, line 2613: Perhaps we should have a newtype or alias for `(ByteString, ByteString)`.

src/Concordium/Client/GRPC2.hs
```diff
@@ -2359,42 +2360,343 @@ instance ToProto SendBlockItemInput where
...
+-- strictly shorter, the last block in the list is the genesis block.
+getAncestorsV2 :: (MonadIO m) => BlockHashInput -> Word64 -> ClientMonad m (GRPCResult (Maybe (Seq.Seq BlockHash)))
+getAncestorsV2 bhInput limit = withServerStreamCollectV2 (callV2 @"getAncestors") msg ((fmap . fmap . mapM) fromProto)
+  where
+    msg = 
+        defMessage
+            & ProtoFields.blockHash .~ toProto bhInput
+            & ProtoFields.amount .~ limit
+
+-- |Get all smart contract modules in a given block.
```
Bargsteen, line 2638: The phrasing used for `getAccountListV2` is more precise and clear. This sounds like it only will return modules from this specific block.

src/Concordium/Client/Runner/Helper.hs
```diff
@@ -39,6 +43,41 @@ type GRPCResult a = Either String (GRPCResponse a)
...
+    \case
+      RawUnaryOutput r ->
+        case r of
+          Right val -> do
+            let (hds, _, response) = val
+            case response of
+              Left e  -> Left $ "gRPC error: " ++ Network.URI.Encode.decode e
+              Right v -> Right (GRPCResponse hds v)
+          Left e -> Left $ "Unable to send consensus query: " ++ show e
+      -- FIXME: There are two sets of headers here; it is not clear to me
```
Bargsteen, line 68: Will you address this is a later PR?

Bargsteen (approved):
Looks good to me apart from the single FIXME. Solid work 👍

src/Concordium/Client/GRPC2.hs
```diff
@@ -2359,42 +2360,343 @@ instance ToProto SendBlockItemInput where
...
+    fromProto ikvPair = do
+        key <- ikvPair ^? ProtoFields.key
+        value <- ikvPair ^? ProtoFields.value
+        return (key, value)
+
+instance FromProto Proto.DelegatorInfo where
+    type Output' Proto.DelegatorInfo = QueryTypes.DelegatorInfo
+    fromProto dInfo = do
+        pdiAccount <- fromProto =<< dInfo ^? ProtoFields.account
+        pdiStake <- fromProto =<< dInfo ^? ProtoFields.stake
```
abizjak, line 2396: So the same comments about using `^?` for lenses apply here as well as on the other PR. Please address those.

src/Concordium/Client/GRPC2.hs
```diff
@@ -2552,20 +2854,255 @@ getAccountInfoV2 ::
...
+        callHelper client = do
+            res <- rawStreamServer method client acc input handler'
+            return $ fmap ServerStreamOutput res
+
+-- |Run a request helper function with a client instance to call a GRPC procedure.
+-- The output is interpreted using the function given in the second parameter.
+--
+-- If the client fails during a call, the connection will be re-established at most
+-- a number of of times depending on the configuration. It should, however, be noted
+-- that for performance reasons, the client is a (global) instance protected by a
```
abizjak, line 2953: What do you mean "for performance reasons"? Performance is not the reason for using the RWLock.

the reason for using it is so that reconnects can be handled since the library we have is shitty and does not properly handle timeouts, among other things.

src/Concordium/Client/Runner/Helper.hs
```diff
@@ -39,6 +43,37 @@ type GRPCResult a = Either String (GRPCResponse a)
...
+outputGRPCV2' =
+    \case
+      RawUnaryOutput r ->
+        case r of
+          Right val -> do
+            let (hds, _, response) = val
+            case response of
+              Left e  -> Left $ "gRPC error: " ++ Network.URI.Encode.decode e
+              Right v -> Right (GRPCResponse hds v)
+          Left e -> Left $ "Unable to send consensus query: " ++ show e
```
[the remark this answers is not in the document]
vikt0r0: Can you please be a little specific about what you want here?

Did you see the documentation for the type `data GRPCOutput`? I think that perfectly suffices.

src/Concordium/Client/GRPC2.hs
vikt0r0: > Performance is not the reason for using the RWLock.

Right. But the message I am trying to convey in that sentence is not about the RWLock, but rather about the single (global) client instance, i.e.:

> for performance reasons, _the client is a (global) instance_

If I am wrong about this, or if this is poorly worded, then the reason should be rephrased or perhaps completely omitted.

src/Concordium/Client/GRPC2.hs
vikt0r0: Thanks.

src/Concordium/Client/GRPC2.hs
vikt0r0: Yes, I actually thought about that as well! However I think that reading the current type signature, it is clear that this is a key value pair. My concern is that by adding an alias/newtype, we simply add another level of abstraction and the user will have to look it up. But let me know what you think :-)

Bargsteen: Yea, you might be right.
I think you should be more explicit about what the return type is in the docs. Then it is fine.

src/Concordium/Client/Runner/Helper.hs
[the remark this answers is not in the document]
abizjak: The thing is, I think you can make this code much easier to read than it is. The comment to the type does not really provide much insight, you have to go to https://hackage.haskell.org/package/http2-client-grpc-0.8.0.0/docs/Network-GRPC-Client.html#t:RawReply but that doesn't explain anything either.

You presumably know now what the response triple is. So you could explain it to a future reader, to make up for the inadequacy of the library documentation.

For example 
```
      -- |The output returned by invoking a GRPC procedure using `rawStreamServer`.
```
does not provide much information to me. You could explain what the triple is.

The second point is that, as per contribution guidelines, you should either make changes based on the feedback, or you should provide a response saying why you will not do what was asked. For example you could have written "There is documentation of the type which I think suffices" or some such.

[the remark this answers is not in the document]
abizjak: Also
```
          Left e -> Left $ "Unable to send consensus query: " ++ show e
```
is not really correct. This is used for other queries than consensus queries I assume. Like getting node info, no?

src/Concordium/Client/GRPC2.hs
abizjak: It's not really a global instance in the normal sense of the word. You are meant to use the same connection for multiple requests since establishing connections is expensive. You should write that since these comments about "global" instance are misleading. As in, I wrote this setup in the client, and I don't understand what you mean by the comment, so I think other readers will likely not either.

src/Concordium/Client/Runner/Helper.hs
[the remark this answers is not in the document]
vikt0r0: >Also

>          Left e -> Left $ "Unable to send consensus query: " ++ show e

>is not really correct. This is used for other queries than consensus queries I assume. Like getting node info, no?

Yes. This is what is in the function, and I am wary about quote-unquote "casting aspersions" towards things and changing things I did not write based on your previous feedback.

I just lifted it into a case-statement. I can remove the "consensus" part of it.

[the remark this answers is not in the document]
vikt0r0: I made the following changes based on your feedback:

- The `data GRPCOutput` declaration literally reads `-- |The output returned by invoking a GRPC procedure using `rawStreamServer` and provides a link for the reader of the comment.

- If you read a few lines ahead, there also was a comment with questions regarding the two sets of headers. I did not know at the time what the headers were, so your presumption is incorrect. Following your answer I did document the headers there, i.e. stating which is the headers respectively trailers.

PR #212, opened 2023-02-13
Fix buffering of streaming calls by updating the dependency.

### LINKED ISSUE #213: Streaming RPC calls are slow if there is more than ~64kB of data
Streaming RPC calls in the client buffer (i.e., hang for a few seconds) after receiving some amount of data. That seems to be about 64kB of data.

The reason for this is that in http2 protocol 

```
   The receiver of a frame sends a WINDOW_UPDATE frame as it consumes
   data and frees up space in flow-control windows.  Separate
   WINDOW_UPDATE frames are sent for the stream- and connection-level
   flow-control windows.
```

that is, both the individual stream **and the connection** must be credited when data is received.

The client (or rather its dependency) only credits the stream, and relies on a background thread which runs every 5s by default.

The observed behaviour is that the response stops being processed for a few seconds after receiving 64kB of data.

This is relevant for the V2 revision which has a number of streaming calls.

abizjak:
## Purpose

Fix delays in streaming calls due to uncredited connection window updates.

Depends on 
- [x] https://github.com/Concordium/http2-grpc-haskell/pull/4

Fixes #213 

## Checklist

      hard-to-understand areas.

### DIFF
--- a/stack.yaml
+++ b/stack.yaml
@@ -45,7 +45,7 @@ extra-deps:
   commit: 8087dfd416118064934affb35caab811a8a70233
 
 - github: Concordium/http2-grpc-haskell
-  commit: 7f4f4955cd64d867b9fda951600a445653f5b098
+  commit: d3d7a3413ccf4ca899602df8ef8e99601024938a
   subdirs:
     - http2-client-grpc
     - http2-grpc-proto-lens
--- a/stack.yaml
+++ b/stack.yaml
@@ -45,7 +45,7 @@ extra-deps:
   commit: 8087dfd416118064934affb35caab811a8a70233
 
 - github: Concordium/http2-grpc-haskell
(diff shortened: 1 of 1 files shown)


vikt0r0 (approved):

stack.yaml
```diff
@@ -45,7 +45,7 @@ extra-deps:
   commit: 8087dfd416118064934affb35caab811a8a70233
 
 - github: Concordium/http2-grpc-haskell
-  commit: 7f4f4955cd64d867b9fda951600a445653f5b098
+  commit: d3d7a3413ccf4ca899602df8ef8e99601024938a
```
abizjak, line 48:
```suggestion
  commit: d79454adc7234908a05b3b5339b48ceaca23ffd9
```

PR #211, opened 2023-02-11
Migrate family of `raw` client commands to GRPC V2.

vikt0r0:
## Purpose

Migrate the `raw` family of commands to use the GRPC V2 API; this is part of migrating the client to the GRPC V2 API feature.

Depends on https://github.com/Concordium/concordium-base/pull/317

## Changes
Implement helpers for gluing V2 API and `raw` commands of the client:

### Tasks
- Wrap output of `fromProto` in `Either` rather than `Maybe`. 
- Migrate `raw` commands.
- Misc. fixes and better comments in GRPC V2 `FromProto` instances.

### Client interface changes
- Remove `raw` client commands `startBaker`, `stopBaker`, `joinNetwork`, `leaveNetwork`.
- Remove node ID input from `BanNode` and `UnbanNode`.
- Add dump file path and `raw` flag to `DumpStart`.
- Change output of `GetBannedPeers` to display JSON instead of using leveraging `show`, e.g.

```
Right {peer_type: "Node" peers { node_id { value: "*" } ip { value: "10.10.10.10" } catchup_status: PENDING } peers { node_id { value: "*" } ip { value: "192.168.0.1" } catchup_status: PENDING } peers { node_id { value: "*" } ip { value: "192.168.0.100" } catchup_status: PENDING } peers { node_id { value: "*" } ip { value: "192.168.1.1" } catchup_status: PENDING } peers { node_id { value: "*" } ip { value: "192.168.1.100" } catchup_status: PENDING }}
```

to:

```
[
    "10.10.10.10",
    "192.168.0.1",
    "192.168.0.100",
    "192.168.1.1",
    "192.168.1.100"
]
```

- Change output of `GetPeerUptime` to display a `Word64` instead of `Maybe Word64`.
- Output of `GetTransactionStatusInBlock` when no transaction summary is present in response.

## Checklist
- [ ] My code follows the style of this project.
- [ ] The code compiles without warnings.
- [ ] I have performed a self-review of the changes.
- [ ] I have documented my code, in particular the intent of the
      hard-to-understand areas.
- [ ] (If necessary) I have updated the CHANGELOG.

PR #214, opened 2023-02-16
Migrate family of `account` client commands to GRPC V2.

vikt0r0:
## Purpose

Migrate the `account` family of commands to use the GRPC V2 API; this is part of migrating the client to the GRPC V2 API feature.

Depends on:

https://github.com/Concordium/concordium-base/pull/317

https://github.com/Concordium/concordium-client/pull/211

## Changes
Implement helpers for gluing V2 API and `account` commands of the client:

### Tasks
- Migrate `account` commands.
- Refactoring due to dependencies.

### Client interface changes
- Different messages in case of API server errors and response conversion/invariant errors.

## Checklist
      hard-to-understand areas.


src/Concordium/Client/GRPC2.hs
```diff
@@ -2533,6 +2536,303 @@ instance FromProto Proto.PendingUpdate where
                     PUETimeParameters <$> fromProto tParameters
         return PendingUpdate{..}
 
+type LoggerMethod = Text -> IO ()
```
vikt0r0, line 2539: Perhaps this module should be split into two, a conversion-related and a client-logic related one. How do you feel about this?

vikt0r0:
In this case, changes were made to error messages. I could use some help figuring out how this should be documented in the changelog? Is it sufficient to mention, provide a few example of how, or document the changes in a somewhat more exhaustive manner?

src/Concordium/Client/GRPC2.hs
```diff
@@ -2533,6 +2536,303 @@ instance FromProto Proto.PendingUpdate where
...
+--
+-- (WriteLocked n, ⊤, ⊥) -AR-> (WriteLocked n, ⊤, ⊥)
+-- (WriteLocked n, ⊤, ⊥) -AW-> (WriteLocked (n+1), ⊤, ⊥), WR
+-- (WriteLocked n, ⊤, ⊥) -RW-> (Free n, ⊤, ⊤), WW
+--
+-- No other state should be reachable.
+--
+-- Additionally, rwlReadLock and rwlWriteLock can only be modified while the
+-- rwlState MVar is held.
+data RWLock = RWLock {
```
Bargsteen, line 2688: Since you are moving this lock, I think you should put it in its own file.

src/Concordium/Client/Runner/Helper.hs
```diff
@@ -132,3 +135,37 @@ value s =
...
+    Right resp ->
+      case grpcResponseVal resp of
+        Left err -> Left $ "Unable to convert response payload: " <> err
+        Right val -> Right $ f val
+
+-- |Extract the response value of a GRPCResult and return it under the
+-- provided mapping or fail with an error message if the result contains
+-- an error. Takes a string to be prepended to the error message.
+extractResponseValueOrFail :: (MonadIO m)
+  => (a -> b) -- |Result mapping
```
Bargsteen, line 157: `-- |` refers to the item below, so this will be shown for the `String` parameter. Move the comment or use ` -- ^ blah` instead 😊

src/Concordium/Client/Types/TransactionStatus.hs
```diff
@@ -25,6 +27,31 @@ data TransactionStatusResult' a = TransactionStatusResult
 
 type TransactionStatusResult = TransactionStatusResult' ValidResult
 
+-- |Convert an `TransactionStatus` instance into a `TransactionStatusResult` instance.
```
Bargsteen, line 30:
```suggestion
-- |Convert a `TransactionStatus` instance into a `TransactionStatusResult` instance.
```

src/Concordium/Client/Types/TransactionStatus.hs
```diff
@@ -25,6 +27,31 @@ data TransactionStatusResult' a = TransactionStatusResult
...
+-- Returns @Left@ wrapping an error message if a transaction summary was @Nothing@ if
+-- the input is either @Committed@ or @Finalized@, or a @Right@ wrapping the result
+-- otherwise.
```
Bargsteen, line 33: Something seems off in this sentence. Please revise it. Perhaps it just missing an `or` after `@Nothing@`, but it could also be made a bit clearer.

src/Concordium/Client/Runner.hs
```diff
@@ -1494,6 +1485,22 @@ tailTransaction verbose hash = do
...
+readOrFail t =
+  case readEither s of
+    Left err -> logFatal [[i|Unable to parse '#{s}': #{err}|]]
+    Right v -> return v
+  where s = Text.unpack t
+
+-- |Reads a blockhash wrapped in the `Maybe` monad.
+-- If the provided value is @Nothing@, a default value specified provided
+-- as the first parameter is returned. If the provided value is string wrapped
+-- in @Just@, `readOrFail s` is returned. Fails if `s` is not a valid blockhash.
```
Bargsteen, line 1499: When you refer to `s` it forces the user of this function to visit the definition. Ideally, you would be able to just read the comment and understand how to use it. Especially for such a simple function.

src/Concordium/Client/Runner.hs
```diff
@@ -2064,14 +2069,9 @@ processContractCmd action baseCfgDir verbose backend =
...
-  res <- getInstanceInfo (Text.pack . showCompactPrettyJSON . ncaAddr $ namedContrAddr) block
-  case res of
-    Left err -> logFatal ["I/O error:", err]
-    -- TODO: Handle nonexisting blocks separately from nonexisting contracts.
-    Right (GRPCResponse _ AE.Null) -> logFatal [[i|the contract instance #{showNamedContractAddress namedContrAddr} does not exist in block #{block}|]]
-    Right contrInfo -> case AE.fromJSON $ grpcResponseVal contrInfo of
-      Error err -> logFatal ["Could not decode contract info:", err]
-      Success info -> pure info
+  b <- readOrFail block
+  r <- getResponseValueOrFail =<< getInstanceInfoV2 (ncaAddr namedContrAddr) (Given b)
```
Bargsteen, line 2073: What is now printed if the contract doesn't exist?
We should make sure to still print a nice message, as that will happen quite frequently.

src/Concordium/Client/Runner.hs
```diff
@@ -2165,14 +2165,9 @@ getWasmModule :: (MonadIO m)
...
-  res <- getModuleSource (Text.pack . show $ nmrRef namedModRef) block
-
-  case res of
-    Left err -> logFatal ["I/O error:", err]
-    Right (GRPCResponse _ "") -> logFatal [[i|the module reference #{showNamedModuleRef namedModRef} does not exist in block #{block}|]]
-    Right unparsedWasmMod -> case S.decode $ grpcResponseVal unparsedWasmMod of
-      Left err' -> logFatal [[i|could not decode Wasm Module:|], err']
-      Right wasmMod -> return wasmMod
+  bh <- readOrFail block
+  getModuleSourceV2 (nmrRef namedModRef) (Given bh) >>=
```
Bargsteen, line 2169: Again, what will be printed to the user when the module doesn't exist?

src/Concordium/Client/Runner.hs
```diff
@@ -3870,10 +3849,10 @@ processTransaction ::
   => BSL.ByteString
   -> Int
   -> ClientMonad m Types.BareBlockItem
-processTransaction source networkId =
+processTransaction source _networkId =
   case AE.eitherDecode source of
     Left err -> fail $ "Error decoding JSON: " ++ err
-    Right t  -> processTransaction_ t networkId True
+    Right t  -> processTransaction_ t _networkId True
```
Bargsteen, line 3855: Why did you make this change?

Bargsteen (commented):
Looks good, but I want to make sure the error messages for common user mistakes remain.

src/Concordium/Client/GRPC2.hs
vikt0r0: Do you want a file which contains this lock, or are you referring to my comment here:

https://github.com/Concordium/concordium-client/pull/214#issuecomment-1433282886

As mentioned I think it makes sense to split the current GRPC2 module :-)

src/Concordium/Client/Runner/Helper.hs
vikt0r0: Thanks. So we use haddock documentation?

Do I read [this](http://haskell-haddock.readthedocs.io) to brush up on comment guidelines?

src/Concordium/Client/Runner.hs
vikt0r0: It prints the node response, in this case it would be:

`Error: A GRPC error occurred: gRPC error: block or instance not found.`

Which is the response from the node. It is not possible to really discern further based on the response.

src/Concordium/Client/Runner.hs
vikt0r0: The node response stating the nature of the error.

src/Concordium/Client/Types/TransactionStatus.hs
vikt0r0: Yea, I should revise the comments. It seems there are some slip-ups here and there. Will do.

src/Concordium/Client/Runner.hs
Bargsteen: Ok, I also saw `_networkId` changes in one or more of the other PRs.

src/Concordium/Client/Runner/Helper.hs
Bargsteen: Right, I guess it is something you pick up over time via reviews.
We don't have any clear guidelines, but extrapolating from our Haskell code is not always the best idea, as it is often lacking in documentation.

But reading a bit about the haddock syntax (even though we don't host an internal haddock yet) is useful.
Besides that, I try to convey things that would be useful for someone calling the function so that they don't need to look at the code.

Currently, I follow this rough template (private functions may naturally have less info)

```
One line that explains what the function does.

Perhaps more details about the functionality.

Preconditions (if any)

What it returns (if not obvious)
 - Including when it will logFatal / panic etc.
```

src/Concordium/Client/GRPC.hs
```diff
@@ -343,203 +240,6 @@ getCryptographicParameters :: MonadIO m => Text -> ClientMonad m (GRPCResult Val
 getCryptographicParameters hash = withUnary (call @"getCryptographicParameters") msg getJSON
   where msg = defMessage & CF.blockHash .~ hash
 
--- |Cookie headers that may be returned by the node in a query.
-type CookieHeaders = Map.Map BS8.ByteString BS8.ByteString
```
abizjak, line 347: I am really not a fan of making these moves as part of a bigger PR since the rest of the module is staying where it was.

This lock could have stayed where it was, and imported from the other module.

Alternatively we could move the lock into a separate module as part of its own PR that does just that. And that would be merged to main directly quickly, which would make it easier to review as well.

src/Concordium/Client/GRPC2.hs
abizjak: If you move the lock into its own module **in its own PR** then I don't think there is much advantage in splitting the module further.

I don't mind it either, but it needs to be done in self-contained pull requests that are simple to review.

src/Concordium/Client/Runner.hs
abizjak: Also, as I commetned in the other PR where you have the same function, please don't use "monad" for this. There is nothing at all "monadic" in this use of Maybe, so it is just needles jargon.

src/Concordium/Client/Runner.hs
abizjak: Well it is. And the error as it was before was just better than it is now. So please keep the error as it was, there is no reason to change it.

Further, you have deleted a `TODO` which you could handle here as well. In the error case you could query whether the block exists (e.g., getBlockInfo), and print the corresponding error message.

src/Concordium/Client/Runner.hs
abizjak: Again, this error is just worse than the one we had. So please make the error at least as good as it was.

src/Concordium/Client/Runner.hs
```diff
@@ -3708,7 +3687,7 @@ printPeerData bootstrapper pInfos Queries.NodeInfo{..} =
       -- Filter bootstrappers.
       pInfos' = filter (\p -> bootstrapper || Queries.consensusInfo p /= Queries.Bootstrapper) pInfos
   in
-  liftIO $ do
+  liftIO $ do
```
abizjak, line 3690: This does not seem like a useful change.

src/Concordium/Client/Runner.hs
abizjak: Why are we not just removing the network ID? It is not used in V2 at all.

src/Concordium/Client/Runner/Helper.hs
```diff
@@ -132,3 +135,37 @@ value s =
...
+    Left err -> logFatal [errPrefix <> err]
+    Right v -> return v
+
+-- |Get the response value of a GRPCResult or fail if the result contains an error.
+getResponseValueOrFail' :: (MonadIO m) => (a -> b) -> GRPCResult (Either String a) -> m b
+getResponseValueOrFail' f = extractResponseValueOrFail f "" 
+
+-- |Get the response value of a GRPCResult or fail if the result contains an error.
+getResponseValueOrFail :: (MonadIO m) => GRPCResult (Either String a) -> m a
+getResponseValueOrFail = extractResponseValueOrFail id ""
```
abizjak, line 171: This needs to be renamed as we discussed in the other PR.

src/Concordium/Client/Types/Contract/Info.hs
```diff
@@ -270,6 +272,31 @@ data ContractStateV0
   | Cs0JSON !AE.Value
   deriving (Eq, Show)
 
+-- |Convert an `InstanceInfo` instance into a `ContractInfo` instance.
```
abizjak, line 275: There is a crucial point that is not mentioned here. namely what is done with the schema.

Please elaborate.

src/Concordium/Client/Types/TransactionStatus.hs
abizjak: Also, this is not Markdown. Use the correct Haddock markup please.

src/Concordium/Client/GRPC.hs
vikt0r0: I agree that this should probably be in a separate PR. As I recall I had a problem when introducing rewriting the `TransactionStatusQuery` instance using the V2 API as it introduced a cyclic dependency. I'll keep this in mind from now on. Is this OK as-is for now?

vikt0r0: Otherwise just let me know and I will make a PR 👍🏼

src/Concordium/Client/GRPC2.hs
vikt0r0: Resolved due to the discussion in another thread.

src/Concordium/Client/Runner.hs
vikt0r0: I added a TODO on Zenhub for this.

src/Concordium/Client/Runner/Helper.hs
vikt0r0: Thanks for taking your time to elaborate on this. I think the template forms a great basis, and I saved it for future reference. I added a TODO on Zenhub which serves to ensure that all added docs follow best-practices. Once this PR (and its dependency in this repository) is merged into the feature branch I will make a pass to ensure this is the case.

src/Concordium/Client/Types/TransactionStatus.hs
vikt0r0: I will have to go through all the comments to ensure they are proper Haddock markup. I added a separate TODO for this, and I will do this for the entire feature branch.

src/Concordium/Client/Types/TransactionStatus.hs
vikt0r0: I added a TODO to the issue.

src/Concordium/Client/Runner/Helper.hs
vikt0r0: Super. I handle this in later commit which I will request review for as soon as this is approved.

src/Concordium/Client/GRPC.hs
abizjak: I would prefer if you make a PR moving the lock to its own module (and exporting only the required functions from there).

src/Concordium/Client/GRPC2.hs
```diff
@@ -3106,21 +3116,21 @@ withGRPCCoreV2 helper k = do
...
+                else return $ k (RequestFailed "Cannot establish connection to GRPC endpoint.")
         (_, Just v) ->
             let response = toGRPCResult' v
              in do
                     addHeaders response
                     return (k response)
   where
     addHeaders response = case response of
-        Right GRPCResponse{..} -> do
+        StatusOk GRPCResponse{..} -> do
```
vikt0r0, line 3127: Should this not be done for non-'OK' error messages as well?

src/Concordium/Client/Runner.hs
```diff
@@ -1385,15 +1430,13 @@ startTransaction txCfg pl confirmNonce maybeAccKeys = do
...
                      Just acKeys' -> return acKeys'
-                     Nothing -> liftIO $ failOnError $ decryptAccountKeyMapInteractive esdKeys (Nothing) Nothing
+                     Nothing -> liftIO $ failOnError $ decryptAccountKeyMapInteractive esdKeys Nothing Nothing
   let sender = applyAlias tcAlias naAddr
   let tx = signEncodedTransaction pl sender energy nonce expiry accountKeyMap
   when (isJust tcAlias) $
       logInfo [[i|Using the alias #{sender} as the sender of the transaction instead of #{naAddr}.|]]
-  sendTransactionToBaker tx defaultNetId >>= \case
-    Left err -> fail err
-    Right (GRPCResponse _ False) -> fail "transaction not accepted by the baker"
```
vikt0r0, line 1395: This is changed. Is it bad to rely on the node response text from the API in this case?

abizjak: No, but I would still state something like
```
transaction not accepted by the baker: {error from the node}
```
to provide context for the user.

src/Concordium/Client/GRPC2.hs
abizjak: If we can, yes.

abizjak:
> In this case, changes were made to error messages. I could use some help figuring out how this should be documented in the changelog? Is it sufficient to either mention this or provide a few example of how it changed or document the changes in a somewhat more exhaustive manner?

We don't have to detail those in the changelog. We should just add a line that some error message have been revised. We do not guarantee exact phrasing of error messages.

src/Concordium/Client/Runner.hs
```diff
@@ -1507,32 +1566,28 @@ processAccountCmd action baseCfgDir verbose backend =
...
-            Left err -> logFatal [err]
-            Right aiv -> case AE.fromJSON $ grpcResponseVal aiv of
-                            AE.Error err -> logFatal ["Cannot decode account info response from the node: " ++ err]
-                            AE.Success Nothing -> logFatal [printf "Account %s does not exist on the chain." $ show accountIdentifier]
-                            AE.Success (Just air) -> return air
+        bhInput <- readBlockHashOrDefault Best block
+        accInfo <- getResponseValueOrDie =<< getAccountInfoV2 accountIdentifier bhInput
         -- derive the address of the account from the the initial credential
-        resolvedAddress <- case Map.lookup (ID.CredentialIndex 0) (Types.aiAccountCredentials accInfo) of
-                            Nothing -> logFatal [printf "No initial credential found for the account identified by '%s'" accountIdentifier ]
```
vikt0r0, line 1534: How come `show` works here, when `AccountIdentifier` is not an instance of `show`?

I even get `No instance for (Show Types.AccountIdentifier)` when I try now? What am I overlooking?

Bargsteen: As I understand it, `printf` doesn't use `Show` but rather `PrintfArg`. Also, it is not typesafe, so it is better to use the string interpolation with `logFatal [[i|No initial credential found for the account identified by '#{accountIdentifier}'|]]`.

src/Concordium/Client/GRPC2.hs
vikt0r0: Adding a TODO on Zenhub for this.

src/Concordium/Client/Runner.hs
vikt0r0: Thanks. I'm not sure I understand how that works, but in any case, using string interpolation was what I wanted. I'll add the `Show` instance then.

src/Concordium/Client/Runner.hs
```diff
@@ -1507,32 +1586,33 @@ processAccountCmd action baseCfgDir verbose backend =
...
-                            AE.Error err -> logFatal ["Cannot decode account info response from the node: " ++ err]
-                            AE.Success Nothing -> logFatal [printf "Account %s does not exist on the chain." $ show accountIdentifier]
-                            AE.Success (Just air) -> return air
+        bhInput <- readBlockHashOrDefault Best block
+        accInfo <- do
+          res <- getAccountInfoV2 accountIdentifier bhInput
+          dieOnRequestFailed ["I/O error:"] res
+          dieOnStatusNotFound [[i|Account does not exist on the chain.|]] res
+          dieOnConversionError ["Cannot decode account info response from the node:"] res
+          getResponseValueOrDie res
```
abizjak, line 1610: Is there a reason for not using `getAccountInfoOrDie` here?

abizjak (commented):
Looks OK overall, there are a few minor issues that should be addressed.

src/Concordium/Client/Runner.hs
```diff
@@ -1092,13 +1160,13 @@ getDelegatorCooldown bs = do
...
-    AE.Error err -> logFatal ["Cannot decode account info response from the node: " ++ err]
-    AE.Success Nothing -> logFatal [printf "Account %s does not exist on the chain." $ show senderAddr]
-    AE.Success (Just air) -> return air
+getAccountInfoOrDie :: (MonadIO m) => ID.AccountAddress -> BlockHashInput -> ClientMonad m Types.AccountInfo
+getAccountInfoOrDie senderAddr bhInput = do
+  res <- getAccountInfoV2 (Types.AccAddress senderAddr) bhInput
+  dieOnRequestFailed ["I/O error:"] res
+  dieOnStatusNotFound [[i|Account #{senderAddr} does not exist on the chain.|]] res
+  dieOnConversionError ["Cannot decode account info response from the node:"] res
+  getResponseValueOrDie res
```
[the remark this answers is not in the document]
vikt0r0: My reasoning for not using the case analysis is that the checks are done in a lot of places, and that I thought it was

1) hard to read (and perhaps quite error prone) as the result datatype is quite complex,
2) the result datatype subject to change (e.g. it has to carry headers whenever the request succeeded at the application level, in a future PR). So if we pattern match everywhere, it think it would be a bit harder to maintain.

This pattern is used a lot in the subsequent PRs, which I will request review on once this is done.

Note that `getResponseValueOrDie` prints an error message as well. The various checks are to keep the error messages as-is as instructed. If you think it is better I can rewrite everything to case expressions, it is like 20 or-so places.

src/Concordium/Client/Runner.hs
```diff
@@ -2165,14 +2248,15 @@ getWasmModule :: (MonadIO m)
...
-      Left err' -> logFatal [[i|could not decode Wasm Module:|], err']
-      Right wasmMod -> return wasmMod
+  bh <- readOrFail block
+  blockRes <- getBlockInfoV2 (Given bh)
+  dieOnRequestFailed ["I/O error:"] blockRes
+  dieOnStatusNotFound [[i|block #{block} does not exist|]] blockRes
+  res <- getModuleSourceV2 (nmrRef namedModRef) (Given bh)
+  dieOnRequestFailed ["I/O error:"] res
+  dieOnStatusNotFound [[i|the module reference #{showNamedModuleRef namedModRef} does not exist in block #{block}|]] res
+  dieOnConversionError ["could not decode Wasm Module:"] res
```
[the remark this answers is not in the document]
vikt0r0: I can change it to case-expressions and get rid of the. If it is the check-and-extract pattern which is ugly, it could perhaps be changed to return `[String] -> GRPCResult a -> Either [String] GRPCResult a`, and then sequenced? Then we can `logFatal` on an error? Or is it because case-expressions are altogether better.

src/Concordium/Client/Runner.hs
vikt0r0: Yea, due to the signature, but that's changed in the next PR.

src/Concordium/Client/Runner/Helper.hs
```diff
@@ -132,3 +164,47 @@ value s =
...
 getValue :: forall a b. (Field.HasField a "value" b) => SimpleGetter (GRPCResponse a) (GRPCResponse b)
 getValue = to (fmap (^. CF.value))
+
+-- |Extract the response value of a @GRPCResult@, if present, and return it
+-- under the provided mapping.
+-- Returns a @Left@ wrapping an error string describing its nature if the
+-- request could not be made, or if the GRPC status code was not 'OK', or a
+-- @Right@ wrapping the response value under the provided mapping otherwise.
+-- VH/FIXME: Change @Either String a@ to @FromProtoResult a @ after re-
+-- factoring.
```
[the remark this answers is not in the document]
vikt0r0: I can not resolve this yet due to cyclic dependencies.

abizjak (approved):

PR #215, opened 2023-02-16
Migrate family of `transaction` client commands to GRPC V2.

vikt0r0:
## Purpose

Migrate the `transaction` family of commands to use the GRPC V2 API; this is part of migrating the client to the GRPC V2 API feature.

Depends on:

https://github.com/Concordium/concordium-base/pull/317

https://github.com/Concordium/concordium-client/pull/214

## Changes
Implement helpers for gluing V2 API and `transaction` commands of the client:

### Tasks
- Migrate `transaction` commands.

### Client interface changes
- Different messages in case of API server errors and response conversion/invariant errors.

## Checklist
- [X] My code follows the style of this project.
- [X] The code compiles without warnings.
- [X] I have performed a self-review of the changes.
- [X] I have documented my code, in particular the intent of the
      hard-to-understand areas.
- [ ] (If necessary) I have updated the CHANGELOG.

PR #222, opened 2023-02-17
Migrate `block`, `identity`, `module`, `contract`, `consensus`, `delegator` and `baker` client commands to GRPC V2.

vikt0r0:
## Purpose

Migrate the `block`, `identity`, `module`, `contract`, `consensus`, `delegator` and `baker` families of commands to use the GRPC V2 API; this is part of migrating the client to the GRPC V2 API feature.

Depends on:

https://github.com/Concordium/concordium-base/pull/317

https://github.com/Concordium/concordium-client/pull/215

## Changes
Implement helpers and make changes to glue together V2 API and commands of the client:

### Tasks
- Migrate `block`, `identity`, `module`, `contract`, `consensus`, `delegator` and `baker` commands.

### Client interface changes
- Slightly different (but hopefully improved) messages in case of API server errors and response conversion/invariant errors.

## Checklist
      hard-to-understand areas.


src/Concordium/Client/Runner.hs
```diff
@@ -3346,15 +3439,15 @@ processDelegatorConfigureCmd baseCfgDir verbose backend txOpts cdCapital cdResta
...
-            Nothing -> [i||]
+      cooldownDate <- withClient backend $ do
+        bcpRes <- getBlockChainParametersV2 Best
+        case getResponseValue bcpRes of
+          Left (_, err) -> do
+            logError ["Could not get the delegator cooldown period: " <> err]
+            exitTransactionCancelled
+          Right v -> do
+            liftIO $ getDelegatorCooldown v
+      let cooldownString :: String = [i|The current baker cooldown would last until approximately #{cooldownDate}|]
```
vikt0r0, line 3450: Should this say "delegator cooldown"?

src/Concordium/Client/Runner.hs
```diff
@@ -1057,61 +1052,62 @@ getEncryptedAmountTransferData senderAddr ettReceiver ettAmount idx secretKey =
...
 getPoolStatusOrDie mbid = do
-  let (bid, passiveDelegation) = case mbid of
-        Nothing -> (0, True)
-        Just bakerId -> (bakerId, False)
-  infoValue <- logFatalOnError =<< withBestBlockHash Nothing (getPoolStatus bid passiveDelegation)
-  case AE.fromJSON $ grpcResponseVal infoValue of
-    AE.Error err -> logFatal ["Cannot decode pool status response from the node: " ++ err]
-    AE.Success Nothing -> logFatal ["Could not query pool status from the chain."]
-    AE.Success (Just ps) -> return ps
+  psRes <- getPoolInfoV2 Best (fromMaybe 0 mbid)
```
abizjak, line 1100: I don't think this is correct. If `Nothing` is supplied you are meant to query passive delegation if I recall correctly.

src/Concordium/Client/Runner.hs
```diff
@@ -1750,19 +1751,40 @@ processModuleCmd action baseCfgDir verbose backend =
...
-                               "the provided block does not exist:", Text.unpack bestBlock]
-        Just [] -> logInfo ["there are no modules in block " ++ Text.unpack bestBlock]
-        Just xs -> runPrinter $ printModuleList (bcModuleNameMap baseCfg) xs
+      -- Verify that the provided blockhash is valid.
+      case block of
+        Nothing -> return ()
+        Just b -> case readEither (Text.unpack b) of
+          Left _ -> logFatal ["Could not retrieve the list of modules",
+                              "the provided block hash is invalid, tried:",
+                              Text.unpack b]
```
abizjak, line 1760: Won't this be on a separate line? Why?

src/Concordium/Client/Runner.hs
```diff
@@ -1959,14 +2014,18 @@ processContractCmd action baseCfgDir verbose backend =
         Just (InvokerAccount nameOrAddr) -> Just . Types.AddressAccount . naAddr <$> getAccountAddressArg (bcAccountNameMap baseCfg) nameOrAddr
         Just InvokerContract{..} -> Just . Types.AddressContract . ncaAddr <$> getNamedContractAddress (bcContractNameMap baseCfg) icIndexOrName icSubindex
 
-      contrInfo <- withClient backend . withBestBlockHash block $ getContractInfo namedContrAddr
+      (bbHash, contrInfo) <- withClient backend $ do
+        bhInput <- readBlockHashOrDefault Best block
+        bHash <- extractResponseValueOrDie Queries.biBlockHash =<< getBlockInfoV2 bhInput
```
abizjak, line 2019: So you are doing this a lot, but it is not necessary. It is not wrong, but each request returns in its header the block hash that was used, so if you supply Best you will get back a header blockHash with the actual hash that was used.

this would save these requests. I don't want you to spend time changing things now, but wanted to bring it up.

src/Concordium/Client/Runner.hs
```diff
@@ -1978,34 +2037,40 @@ processContractCmd action baseCfgDir verbose backend =
...
-              let eventsMsg = case mapMaybe (fmap (("  - " <>) . Text.pack) . showEvent verbose Nothing) rcrEvents of
-                                [] -> Text.empty
-                                evts -> [i|- Events:\n#{Text.intercalate "\n" evts}|]
-              returnValueMsg <- mkReturnValueMsg rcrReturnValue schemaFile modSchema contractName updatedReceiveName False
-              logSuccess [[iii|Invocation resulted in success:\n
-                               - Energy used: #{showNrg rcrUsedEnergy}
-                               #{returnValueMsg}
-                               #{eventsMsg}
-                               |]]
+      -- VH/FIXME: Should the blockhash input in the following be `Best` or `Given bbHash`?
```
abizjak, line 2041: Yes, it should be the given block. That's the most consistent, even if it was different before.

src/Concordium/Client/Runner.hs
```diff
@@ -3111,12 +3214,14 @@ processBakerUpdateStakeBeforeP4Cmd baseCfgDir verbose backend txOpts ubsStake =
...
-          logError ["Could not reach the node to get the baker cooldown period."]
-          exitTransactionCancelled
+      cooldownDate <- withClient backend $ do
+        bcpRes <- getBlockChainParametersV2 Best
+        case getResponseValue bcpRes of
+          Left (_, err) -> do
+            logError ["Could not get the baker cooldown period: " <> err]
+            exitTransactionCancelled
+          Right v -> do
+            getBakerCooldown v
```
abizjak, line 3224:
```suggestion
          Right v -> getBakerCooldown v
```

src/Concordium/Client/Runner.hs
abizjak: Yes.

src/Concordium/Client/Runner.hs
```diff
@@ -3498,31 +3605,31 @@ processLegacyCmd action backend =
...
-        printResponseValueAsJSON
+          getBlockPendingUpdatesV2 >>=
+            printResponseValueAsJSON
     GetBlockSpecialEvents block ->
       withClient backend $
         readBlockHashOrDefault Best block >>=
-        getBlockSpecialEventsV2 >>=
-        printResponseValueAsJSON
+          getBlockSpecialEventsV2 >>=
+            printResponseValueAsJSON
```
abizjak, line 3622: Why were these changes necessary? Please avoid noise.

src/Concordium/Client/Runner.hs
```diff
@@ -3951,18 +4058,23 @@ processCredential ::
   => BSL.ByteString
   -> Int
   -> ClientMonad m Types.BareBlockItem
-processCredential source networkId =
+processCredential source _networkId =
```
abizjak, line 4061: Why is the parameter not getting removed alltogether?

abizjak (approved):
Please address the issues. Otherwise looks ok.

src/Concordium/Client/Runner.hs
vikt0r0: Thanks for pointing that out. I will have to look into that a bit further.

src/Concordium/Client/Runner.hs
vikt0r0: Yes, I do not know either. Probably it should just be a single line like everything else?

src/Concordium/Client/Runner.hs
vikt0r0: Neat. I'll change it.

src/Concordium/Client/Runner.hs
vikt0r0: 👍🏼

src/Concordium/Client/Runner.hs
vikt0r0: They were not strictly, but I did not know how sequencing should be formatted, and this was consistent with how I ended up doing things elsewhere.

vikt0r0: Do I revert this now?

src/Concordium/Client/Runner.hs
vikt0r0: It wa tempting to do, but I didn't know if this was for a separate PR, so I was a bit wary.

I'll just do it here then?

I do not remember this is the case, or if I perhaps already removed it, but should it be removed from the CLI if it is  there?

src/Concordium/Client/Runner.hs
abizjak: As I wrote, if this is going to delay merging this PR then do not do it.

src/Concordium/Client/Runner.hs
abizjak: Yes.

src/Concordium/Client/Runner.hs
abizjak: We just used default 100 I think. This parameter was never used even in the old API.

src/Concordium/Client/Runner.hs
vikt0r0: fourmolu is not used in the client is it? I recall using it on a file in the client and it pretty much changed up everything.

Is what you are proposing that I use fourmolu config of concordium-base to see how it formats things (in this case sequencing) and then go from that that as a reference?

src/Concordium/Client/Runner.hs
vikt0r0: I don't think it will, but if it seems to be cumbersome I'll leave it as is.

vikt0r0: I gave it a shot, but it took too long, so I opted not to.

PR #226, opened 2023-03-03
Refactor client `RWLock` into separate module, remove GRPC V1 API and middleware and migrate `generator`.

vikt0r0:
## Purpose
As stated in the title.

## Changes
As stated in the title.

## Checklist
      hard-to-understand areas.


vikt0r0:
Is the removal of the middleware considered a "visible" change of `concordium-client`? In that case what goes in the changelog?

src/Concordium/Client/RWLock.hs
```diff
@@ -0,0 +1,201 @@
+module Concordium.Client.RWLock
+  ( RWLock(..)
```
abizjak, line 2: Do we need to export the fields? We should not need that, so unless I am missing something export just the type.

abizjak (approved):
The following files still mention the middleware and we need to remove that since otherwise builds will fail

- ./scripts/distributables/macOS-package/build.sh:153
- ./scripts/distributables/windows-concordium-client.Jenkinsfile
- ./stack.linux-static.yaml

PR #263, opened 2023-06-19
Support grpc updates

eb-concordium:
## Purpose
Closes https://github.com/Concordium/concordium-client/issues/262.

## Changes
* `FromProto` instances have been updated to reflect the changes to `ConsensusStatus`, `BlockInfo` and `BlockBirkParameters`.
* Changes to `printConsensusStatus`:
  - Only print slot duration when present. 
  - Print the newly introduced current timeout duration, current round, current epoch and trigger block time when present.
 * Changes to `printBlockInfo`:
   - Only print slot when present.
   - Print the newly introduced round and epoch when present.
   - Print "Block time" instead of "Slot time".

## Checklist

      hard-to-understand areas.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -8,6 +8,13 @@
 - The client needs node version at least 5.4.
 - Add `raw GetNextUpdateSequenceNumbers` subcommand.
 - Add node version to the output of `raw GetNodeInfo`.
+- Add current timeout duration, current round, current epoch and trigger block time to the output
+  of `raw GetConsensusInfo`.
+- Add current timeout duration, current round, current epoch and trigger block time to the output
+  of `consensus status` when they are present.
+- Add round and epoch to the output of `raw GetBlockInfo`.
+- Add round and epoch to the output of `block show` when they are present.
+- Print "Block time" instead of "Slot time" in the output of `block show`.
 
 ## 5.2.0
 
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -15,6 +15,7 @@
 - Add round and epoch to the output of `raw GetBlockInfo`.
 - Add round and epoch to the output of `block show` when they are present.
 - Print "Block time" instead of "Slot time" in the output of `block show`.
+- In the output of `consensus show-parameters`, only print election difficulty when present. 
 
 ## 5.2.0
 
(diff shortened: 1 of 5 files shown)


src/Concordium/Client/GRPC2.hs
```diff
@@ -843,12 +843,24 @@ instance FromProto Proto.ConsensusInfo where
         csGenesisIndex <- fromProto $ ci ^. ProtoFields.genesisIndex
         csCurrentEraGenesisBlock <- fromProto $ ci ^. ProtoFields.currentEraGenesisBlock
         csCurrentEraGenesisTime <- fmap timestampToUTCTime . fromProto $ ci ^. ProtoFields.currentEraGenesisTime
+        currentTimeoutDuration <- fromProtoMaybe $ ci ^. ProtoFields.maybe'currentTimeoutDuration
+        currentRound <- fromProtoMaybe $ ci ^. ProtoFields.maybe'currentRound
+        currentEpoch <- fromProtoMaybe $ ci ^. ProtoFields.maybe'currentEpoch
+        triggerBlockTime <- (fmap . fmap) timestampToUTCTime . fromProtoMaybe $ ci ^. ProtoFields.maybe'triggerBlockTime
+        let csConcordiumBFTStatus = case (currentTimeoutDuration, currentRound, currentEpoch, triggerBlockTime) of 
+                (Just cbftsCurrentTimeoutDuration, Just cbftsCurrentRound, Just cbftsCurrentEpoch, Just cbftsTriggerBlockTime) -> Just ConcordiumBFTStatus{..}
+                _ -> Nothing
```
abizjak, line 852: I suppose technically we could raise an exception here if only some are set. But I guess we should trust the node and this'll work fine.

ChangeLog.md
```diff
@@ -8,6 +8,13 @@
 - The client needs node version at least 5.4.
 - Add `raw GetNextUpdateSequenceNumbers` subcommand.
 - Add node version to the output of `raw GetNodeInfo`.
+- Add current timeout duration, current round, current epoch and trigger block time to the output
+  of `raw GetConsensusInfo`.
+- Add current timeout duration, current round, current epoch and trigger block time to the output
+  of `consensus status` when they are present.
+- Add round and epoch to the output of `raw GetBlockInfo`.
+- Add round and epoch to the output of `block show` when they are present.
+- Print "Block time" instead of "Slot time" in the output of `block show`.
```
abizjak, line 17: What about election difficulty as part of birk parameters? That's also printed as part of `printQueryBirkParameters` and is no longer present.

abizjak (changes requested):
Looks good for consensus and block info, but I think birk parameters also need to be updated.

ChangeLog.md
abizjak: This was added, so resolving.

abizjak (approved):

EmilLa1 (approved):

PR #270, opened 2023-08-07
Stop (early) consuming a stream if an error was returned.

EmilLa1:
## Purpose

Fix a bug where if a stream response returned an error but yielded no elements, then the client would end up hanging.

This PR leverages the fix here https://github.com/Concordium/http2-grpc-haskell/pull/6 by checking whether the output of the stream returned an error before consuming. 

## Changes

- Update the http2-grpc-haskell dependency that makes this bugfix possible. https://github.com/Concordium/http2-grpc-haskell/pull/6
- Check for the header status code is not 'ok' and return an appropriate error if so. 

## Checklist

      hard-to-understand areas.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- End stream consumption early if an error is returned.
+
 ## 6.0.1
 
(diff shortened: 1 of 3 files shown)


abizjak (approved):
Looks good, seems to work well.

PR #271, opened 2023-08-08
Baker earliest win time queries

td202:
## Purpose

This adds support for getting the earliest projected win time for a baker.

## Changes

- Add `raw GetBakerEarliestWinTime` which gets the raw timestamp for a baker by the baker ID.
- Add `baker win-time` which displays the earliest win time for a baker, with options for displaying in the local timezone (as opposed to UTC) and for polling in a loop.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/src/Concordium/Client/Commands.hs
+++ b/src/Concordium/Client/Commands.hs
@@ -510,6 +510,11 @@ data BakerCmd
         { bodsOpenForDelegation :: !OpenStatus,
           bodsTransactionOpts :: !(TransactionOpts (Maybe Energy))
         }
+    | BakerGetEarliestWinTime
+        { bgewtBakerId :: !BakerId,
+          bgewtShowLocalTime :: !Bool,
+          bgewtPoll :: !Bool
+        }
     deriving (Show)
 
 data DelegatorCmd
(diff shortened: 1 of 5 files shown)


abizjak (approved):
Please list the new commands in the changelog.

src/Concordium/Client/Commands.hs
```diff
@@ -1767,6 +1773,25 @@ bakerUpdateOpenDelegationStatus =
...
+        ( info
+            ( BakerGetEarliestWinTime
+                <$> argument auto (metavar "BAKER-ID" <> help "Baker ID to query.")
+                <*> switch
+                    ( long "local-time"
+                        <> help "Display time in the local time zone (instead of UTC)."
+                    )
+                <*> switch
+                    ( long "poll"
+                        <> help "Repeatedly poll for the latest time."
```
EmilLa1, line 1789: What is the polling interval? Is it relevant for the user? Should it be mentioned?

EmilLa1: Should it be configurable?

abizjak: I think the way it works now does not need any configuration for the polling interval.

EmilLa1 (approved):

src/Concordium/Client/Commands.hs
td202: The interval is variable: 10 seconds if the last reported time was more than 10 seconds in the future, and 1 second otherwise. I don't think there's a need to communicate this or configure it, as this should be fine for most purposes.

Issue #253, opened 2023-04-19
Remove hiding of options via SHOW_ALL_OPTS


abizjak:
**Description**

Currently the client hides a few options behind SHOW_ALL_OPTS environment variable. This was originally done to not confuse users, but the options that are hidden are now documented and sometimes needed. So we should remove this hiding and remove the use of SHOW_ALL_OPTS.

abizjak:
Fixed by https://github.com/Concordium/concordium-client/pull/277

PR #277, opened 2023-09-19
Show the format flag when importing an account

EmilLa1:
## Purpose

Address https://github.com/Concordium/concordium-client/issues/253

By default show the `format` option when importing an account. 

This change will make it easier to import accounts generated as part of genesis creation. 
Normally this flag is not used by users since they get their account by going through an idp via one of the wallets.

However there will soon be a tool available for running a local chain with a custom genesis. 
In this scenario it's desirable to make it easy for users to import their generated accounts into the client so the client 
is easy to use with their new custom chain.

Note also that the default value `mobile` is the same as before, so the user does not have to explicitly provide this flag if importing a mobile backup.

## Changes

Remove the `SHOW_ALL_OPTS` env variable and by default show all possible options at all times.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -11,8 +11,8 @@
   - `GetBakerEarliestWinTime`
   - `GetWinningBakersEpoch`
   - `GetFirstBlockEpoch`
-  - Add support for `CommissionRates` in `CurrentPaydayBakerPoolStatus` (Only    available for node versions > 6.0).
-  - By default show all options for importing an account.
+  - Add support for `CommissionRates` in `CurrentPaydayBakerPoolStatus` (Only available for node versions > 6.0).
+  - Show all options for importing an account.
 
 
 ## 6.0.1
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -11,7 +11,9 @@
   - `GetBakerEarliestWinTime`
   - `GetWinningBakersEpoch`
   - `GetFirstBlockEpoch`
-- Add support for `CommissionRates` in `CurrentPaydayBakerPoolStatus` (Only available for node versions > 6.0).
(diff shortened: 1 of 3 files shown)


ChangeLog.md
```diff
@@ -11,7 +11,9 @@
   - `GetBakerEarliestWinTime`
   - `GetWinningBakersEpoch`
   - `GetFirstBlockEpoch`
-- Add support for `CommissionRates` in `CurrentPaydayBakerPoolStatus` (Only available for node versions > 6.0).
+  - Add support for `CommissionRates` in `CurrentPaydayBakerPoolStatus` (Only    available for node versions > 6.0).
+  - By default show all options for importing an account.
```
abizjak, line 15: There is no "default" anymore. it is not possible to disable this behaviour.

abizjak (commented):
Please link to issue https://github.com/Concordium/concordium-client/issues/253

But note that this isn't really solving the issue. Users don't read command line documentation.

ChangeLog.md
```diff
@@ -11,7 +11,9 @@
   - `GetBakerEarliestWinTime`
   - `GetWinningBakersEpoch`
   - `GetFirstBlockEpoch`
-- Add support for `CommissionRates` in `CurrentPaydayBakerPoolStatus` (Only available for node versions > 6.0).
+  - Add support for `CommissionRates` in `CurrentPaydayBakerPoolStatus` (Only    available for node versions > 6.0).
```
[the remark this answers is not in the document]
EmilLa1: I don't know :shrug:

ChangeLog.md
EmilLa1: Good point.

EmilLa1:
> Please link to issue #253
> 
> But note that this isn't really solving the issue. Users don't read command line documentation.

Nice. Thanks. I somehow missed that issue.

abizjak (approved):

Issue #195, opened 2022-12-13
Display improvements


DOBEN:
**Description**
- [x] Singular use if only one event is found, e.g.
./concordium-client transaction status b6cd9cb0c281d90cd5e9e82f1171b7b6ca7ec6bab12e2b9d48d4224e91434faf (testnet)

- [ ] Add dot after the expression since all other expressions also have a dot, e.g. 
./concordium-client transaction status fcdfd6fbcb2e1b47d22918edf8f701805a92b84cdb647d4061480cf765a8c381 (testnet)
![Screenshot from 2022-12-13 14-23-57](https://user-images.githubusercontent.com/13039169/207369218-6f885b84-0027-4ea3-b7d1-3389fb073009.png)
![Screenshot from 2022-12-13 14-46-40](https://user-images.githubusercontent.com/13039169/207369246-64a0df03-bef2-466e-9b1c-c590049365b5.png)

DOBEN:
Regarding the second bulletpoint, there is a function that might help streamline expressions/punctuation but the initial investigation seems there is more work to do to make expressions/punctuation consistently.
```
-- | Ensure that a string is printable as a sentence by converting the first letter to upper case
--  and, unless it already ends with "standard" punctuation, appending the provided punctionation.
prettyMsg :: String -> String -> String
prettyMsg punctuation = \case
    "" -> ""
    (x : xs) ->
        let s =
                if null xs || last xs `elem` p
                    then xs
                    else xs ++ punctuation
        in  (C.toUpper x) : s
  where
    p = ".,:;?!{}" :: String
```

PR #295, opened 2024-04-11
Fix typos

DOBEN:
## Purpose

addresses https://github.com/Concordium/concordium-client/issues/195

## Changes

- Fix accounting of correctly parsed events


test/SimpleClientTests/TransactionSpec.hs
```diff
@@ -250,16 +250,16 @@ printTransactionStatusTests = describe "print transaction status" $ do
...
-                             "  * transferred 0.000010 CCD from account '2zR4h351M1bqhrL9UywsbHrP3ucA1xY3TBTFRuTsRout8JnLD6' to account '4MkK65HrYvMauNTHTuL23wRDKp4VXkCiTpmoWYFtsrZHV3WwSa'"
+                             "  * transferred 0.000010 CCD from account '2zR4h351M1bqhrL9UywsbHrP3ucA1xY3TBTFRuTsRout8JnLD6' to account '4MkK65HrYvMauNTHTuL23wRDKp4VXkCiTpmoWYFtsrZHV3WwSa'."
                            ]
     describe "committed into two blocks" $
         specify "correct output" $
             p committedSuccessfulAndFailureOutcomes
                 `shouldBe` [ "Transaction is committed into 2 blocks:",
                              "- 0a5d64f644461d95315a781475b83f723f74d1c21542bd4f3e234d6173374389 with status \"success\" and cost 0.000010 CCD (10 NRG):",
-                             "  * transferred 0.000010 CCD from account '2zR4h351M1bqhrL9UywsbHrP3ucA1xY3TBTFRuTsRout8JnLD6' to account '4MkK65HrYvMauNTHTuL23wRDKp4VXkCiTpmoWYFtsrZHV3WwSa'",
+                             "  * transferred 0.000010 CCD from account '2zR4h351M1bqhrL9UywsbHrP3ucA1xY3TBTFRuTsRout8JnLD6' to account '4MkK65HrYvMauNTHTuL23wRDKp4VXkCiTpmoWYFtsrZHV3WwSa'.",
```
abizjak, line 262: Why is this the better output? This is a list so I think it reads better without the stop at the end.

src/Concordium/Client/Output.hs
```diff
@@ -904,10 +904,10 @@ showEvent verbose ciM = \case
     showLoggedEvents :: [Wasm.ContractEvent] -> String
     showLoggedEvents [] = "No contract events were emitted."
     showLoggedEvents evs =
-        [i|#{length evs} contract events were emitted|]
+        [i|#{length evs} contract #{if length evs > 1 then "events were" else ("event was" :: String)} emitted|]
             <> ( if isNothing eventSchemaM
                     then [i| but no event schema was provided nor found in the contract module. |]
-                    else [i|, of which #{length $ filter isLeft $ map showContractEvent evs} were succesfully parsed. |]
+                    else [i|, of which #{length $ filter isRight $ map showContractEvent evs} #{if (length $ filter isRight $ map showContractEvent evs) > 1 then "were" else ("was" :: String)} successfully parsed. |]
```
td202, line 910: Suggestion: factor out `length $ filter isRight $ map showContractEvent evs`.

DOBEN:
A better approach to use this function from the code to convert messages without capitalization and punctuation to ones with:

-- | Ensure that a string is printable as a sentence by converting the first letter to upper case
--  and, unless it already ends with "standard" punctuation, appending the provided punctionation.
prettyMsg :: String -> String -> String
prettyMsg punctuation = \case
    "" -> ""
    (x : xs) ->
        let s =
                if null xs || last xs `elem` p
                    then xs
                    else xs ++ punctuation
        in  (C.toUpper x) : s
  where
    p = ".,:;?!{}" :: String

rasmus-kirk:
A transaction with more than 1 event:
- `d59ad4fc140d1e5e56d551cc73326e3210c0873a4f12550f399a6fa52eb84362`

test/SimpleClientTests/TransactionSpec.hs
DOBEN: Update:
Typo changes will not be addressed in this PR anymore (the second subtask of issue #195 is still open). 
Closing this command as a result.

src/Concordium/Client/Output.hs
```diff
@@ -912,14 +911,16 @@ showEvent verbose ciM = \case
     showLoggedEvents :: [Wasm.ContractEvent] -> String
     showLoggedEvents [] = "No contract events were emitted."
     showLoggedEvents evs =
-        [i|#{length evs} contract events were emitted|]
+        [i|#{length evs} contract #{if length evs > 1 then "events were" else ("event was" :: String)} emitted|]
             <> ( if isNothing eventSchemaM
                     then [i| but no event schema was provided nor found in the contract module. |]
-                    else [i|, of which #{length $ filter isLeft $ map showContractEvent evs} were succesfully parsed. |]
+                    else [i|, of which #{noOfParsedEvents} #{if noOfParsedEvents > 1 then "were" else ("was" :: String)} successfully parsed. |]
```
td202, line 917: Suggestion: instead of `> 1`, use `/= 1` so that the `0` case takes the plural verb.

td202 (approved):

src/Concordium/Client/Output.hs
rasmus-kirk: The 0 case is handled above:

```hs
  showLoggedEvents [] = "No contract events were emitted."
```

But I still think `/= 1` is better.

Issue #296, opened 2024-04-12
Multi-sig signing of ContractUpdate transactions


DOBEN:
**Task description**

Concordium has support for native multi-sig accounts. The goal is to make this feature easily accessible. In particular, person A should be able to sign (not sign and send) a transaction with `concordium-client` and person B should be able to append/pass-in its own signature when signing_and_sending a transaction on-chain.

Using the multi-sig accounts feature via SDK is currently possible (a bit hacky):
https://gist.github.com/DOBEN/683fe1a7c82a0551546a7ec242d30cc0

Related to:
https://github.com/Concordium/concordium-client/issues/38

DOBEN:
I suggest adding an optional ``--no-submit`` flag to toggle between sending or NOT sending the transaction on-chain:
```
concordium-client contract update --signers 0:1,0:2,0:3 --tx-out output.bin/json --no-submit
```

Alternativly, adding a new `sign` command to all transaction types that will NOT send the transaction on-chain would also work:

```
concordium-client contract sign --signers 0:1,0:2,0:3 --tx-out output.bin/json
```

The remaing flow would be:
Adding additional signatures with
```
concordium-client transaction add-signature output.bin --signer 0:2 --tx-out signed-tx.bin/signed-tx.json
```

Submitting the fully signed transaction on chain:
```
```
concordium-client transaction submit signed-tx.bin
```

**Comment by abizjak** [comment]
So the issue is that you will then need multiple "sign" commands, because the transactions are not all under the "transaction" command, e.g., contract updates are under contract, baker stuff is under baker. I don't think this is a good solution.

We already have almost all the options with the existing commands, so it's better to refine it. It'll also be much more straightforward to implement because all "transaction" command share the same command line parser and options.

**Comment by DOBEN** [comment]
The introduction of the `sign` command was an alternative solution to avoid using an extra ` --no-submit` flag but it would mean a lot more code (since we need to add an extra subcommand `sign` for every type of transaction that we have). To simplify the task, let's go with the ``--no-submit`` flag:

The current flow would be:

***Step 1:***

Signig the transaction but not sending it on-chain
```
concordium-client contract update --signers 0:1,0:2,0:3 --tx-out output.bin/json --no-submit
```

***Step 2:***

Adding additional signatures with
```
concordium-client transaction add-signature output.bin --signer 0:2 --tx-out signed-tx.bin/signed-tx.json
```

***Step 3:***

Submitting the fully signed transaction on-chain:
```
[...]

PR #300, opened 2024-05-02
Add multi-sig support

DOBEN:
## Purpose

Addresses https://github.com/Concordium/concordium-client/issues/296
Related https://github.com/Concordium/concordium-base/pull/529

 **The following transaction types were tested using multi-sig:**

- `DeployModule`, `InitContract`, `Update`, `Transfer`, `TransferWithMemo`, `TransferWithScheduleAndMemo`, `TransferWithSchedule`, `ConfigureDelegation`, `ConfigureBaker`, `RegisterData` and `UpdateCredentialKeys`

**The following transaction types were NOT tested using multi-sig:**

- `AddBaker`,  `RemoveBaker` , `UpdateBakerStake`, `UpdateBakerRestakeEarnings`, and `UpdateBakerKeys` (concordium-client created these types in protocol versions below `Types.P4` but not anymore).
https://github.com/Concordium/concordium-client/blob/main/src/Concordium/Client/Runner.hs#L3698
- `UpdateCredentials` (access to ledger + desktop wallet is required for creating a valid credential file - postponed/blocked but planned to be tested later)
- `TransferToEncrypted`, `TransferToPublic`, `EncryptedAmountTransfer` and `EncryptedAmountTransferWithMemo`
(shielded transactions are faced out and the feature will be removed, so the multi-sig feature will not be tested/implemented for it).

**How to test this PR:**

Use these two guides to create a multi-sig account. 
https://gist.github.com/DOBEN/683fe1a7c82a0551546a7ec242d30cc0
https://gist.github.com/limemloh/8c0c55f67cf5a83ac7cc21cb646e65c1

There are two main steps to complete:
- Execute an `update-key` transaction on-chain to associate an additional key to your credential/account + set the key threshold to 2 (alternatively you can also add an additional credential with some keys)
```
concordium-client account update-keys --credId 97f325c...722454df --sender 4jxvYasaPncfmCFCLZCvuL5cZuvR5HAQezCHZH7ZA7AGsRYpix ./update-keys.json  --grpc-port 20000 --grpc-ip node.testnet.concordium.com
```
- Configure `concordium-client` to locally have access to these two keys.
```
concordium-client config account update-keys --keys KEYFILE --account 4jxvYasaPncfmCFCLZCvuL5cZuvR5HAQezCHZH7ZA7AGsRYpix 
```

**Example transaction flow:**
1. Step: Create a ContractUpdate transaction - Sign it with one key - Dump the partially-signed transaction into a `transaction.json` file:
```
stack run concordium-client -- contract update 3383 --signers "0:0" --entrypoint updateOperator --outFile ./transaction.json --parameter-json updateOperator.json --energy 5000 --sender 4jxvYasaPncfmCFCLZCvuL5cZuvR5HAQezCHZH7ZA7AGsRYpix --grpc-port 20000 --grpc-ip node.testnet.concordium.com
```

2. Step: Add an additional signature

```
stack run concordium-client -- transaction add-signature ./transaction.json --signers "0:1" --grpc-port 20000 --grpc-ip node.testnet.concordium.com
```

3. Step: Send the fully signed transaction on-chain

```
stack run concordium-client -- transaction submit ./transaction.json  --grpc-port 20000 --grpc-ip node.testnet.concordium.com
```

4. Alternative Step: Pass in a key file when signing instead of using local keys:
```
stack run concordium-client -- transaction add-signature ./transaction.json --keys ../keypair1.json --grpc-port 20000 --grpc-ip node.testnet.concordium.com
```

## Changes

- Rename subcommand `TransactionSubmit` to `TransactionSignAndSubmit` (keep its functionality)

- Create two new subcommands:
  - `TransactionSubmit`
  - `TransactionAddSignature`
  
- Add optional `--outFile` flag to all transaction-creating commands.


src/Concordium/Client/Runner.hs
```diff
@@ -3472,6 +3653,8 @@ processBakerRemoveCmd baseCfgDir verbose backend txOpts = do
         let payload = Types.encodePayload Types.RemoveBaker
             nrgCost _ = return . Just $ bakerRemoveEnergyCost $ Types.payloadSize payload
         txCfg@TransactionConfig{..} <- getTransactionCfg baseCfg txOpts nrgCost
+        logInfo ["payloadpayloadpayloadpayloadpayloadpayload: "]
```
eb-concordium, line 3656: Is this string intented?

src/Concordium/Client/Commands.hs
```diff
@@ -256,13 +267,16 @@ data AccountCmd
           adAmount :: !Amount,
           -- | Which indices of incoming amounts to use as inputs.
           -- If none are provided all existing ones will be used.
-          adIndex :: !(Maybe Int)
+          adIndex :: !(Maybe Int),
+          -- | Optional file name and path to ouptput the signed/partially-signed transaction to instead of submitting the transaction on-chain.
+          adOutFile :: !(Maybe FilePath)
```
eb-concordium, line 272: Same for this one, already in `TransactionOpts`

src/Concordium/Client/Commands.hs
```diff
@@ -256,13 +267,16 @@ data AccountCmd
...
         }
     | -- | Updated credentials and account threshold (i.e. how many credential holders that need to sign transactions)
       AccountUpdateCredentials
         { aucNewCredInfos :: !(Maybe FilePath), -- File containing the new CredentialDeploymentInformation's
           aucRemoveCredIds :: !(Maybe FilePath), -- File containing the CredentialRegistrationID's for the credentials to be removed
           aucNewThreshold :: !AccountThreshold, -- The new account threshold
+          aucOutFile :: !(Maybe FilePath), -- Otional file name and path to ouptput the signed/partially-signed transaction to instead of submitting the transaction on-chain.
```
eb-concordium, line 279: Same for this one, already in `TransactionOpts`

src/Concordium/Client/Commands.hs
```diff
@@ -913,6 +1002,7 @@ accountEncryptCmd =
             ( AccountEncrypt
                 <$> transactionOptsParser
                 <*> option (eitherReader amountFromStringInform) (long "amount" <> metavar "CCD-AMOUNT" <> help "The amount to transfer to shielded balance.")
+                <*> optional (strOption (long "outFile" <> metavar "FILE" <> help "An optional file name and path to ouptput the signed/partially-signed transaction to instead of submitting the transaction on-chain."))
```
eb-concordium, line 1005: I think this is not needed since it is already in `transactionOptsParserBuilder` used in `transactionOptsParser` :)

src/Concordium/Client/Commands.hs
```diff
@@ -926,6 +1016,7 @@ accountDecryptCmd =
                 <$> transactionOptsParser
                 <*> option (maybeReader amountFromString) (long "amount" <> metavar "CCD-AMOUNT" <> help "The amount to transfer to public balance.")
                 <*> optional (option auto (long "index" <> metavar "INDEX" <> help "Optionally specify the index up to which shielded amounts should be combined."))
+                <*> optional (strOption (long "outFile" <> metavar "FILE" <> help "An optional file name and path to ouptput the signed/partially-signed transaction to instead of submitting the transaction on-chain."))
```
eb-concordium, line 1019: I think this is not needed since it is already in `transactionOptsParserBuilder` used in `transactionOptsParser` :)

src/Concordium/Client/Commands.hs
```diff
@@ -967,6 +1058,7 @@ accountUpdateCredentialsCmd =
                 <$> optional (strOption (long "new-credentials" <> metavar "FILE" <> help "File containing the new credential deployment informations."))
                 <*> optional (strOption (long "remove-credentials" <> metavar "FILE" <> help "File containing credential registration ids of the credentials to be removed."))
                 <*> option auto (long "new-threshold" <> metavar "THRESHOLD" <> help "New account threshold, i.e. how many credential holders needed to sign a transaction.")
+                <*> optional (strOption (long "outFile" <> metavar "FILE" <> help "An optional file name and path to ouptput the signed/partially-signed transaction to instead of submitting the transaction on-chain."))
```
eb-concordium, line 1061: I think this is not needed since it is already in `transactionOptsParserBuilder` used in `transactionOptsParser` :)

eb-concordium (approved):
Looks good :)
Regarding `UpdateCredentials`, I do not agree that access to ledger + desktop wallet is required for creating a valid credential file. You can do that with the tool `client` in `concordium-base/rust-bins/src/bin/` :)

td202 (approved):
Looks good. I think it would be nice for submitting transactions to have better error handling (or perhaps checking before-hand) for the transaction having insufficient signatures or having already expired.

I also wonder if we can just completely remove the "sign-and-submit" behaviour. Part of the problem with it is that it uses a different format from "submit" and the other transaction commands, so it could cause issues for people.

It could also be nice to have an easy way to create a transaction without signing at all. (It is possible by specifying a file with contents `{}` to the `--keys`, so I don't think this is essential to add.)

DOBEN:
> Looks good. I think it would be nice for submitting transactions to have better error handling (or perhaps checking before-hand) for the transaction having insufficient signatures or having already expired.
> 
> I also wonder if we can just completely remove the "sign-and-submit" behaviour. Part of the problem with it is that it uses a different format from "submit" and the other transaction commands, so it could cause issues for people.
> 
> It could also be nice to have an easy way to create a transaction without signing at all. (It is possible by specifying a file with contents `{}` to the `--keys`, so I don't think this is essential to add.)

Great input:

I created an extra issue for checking for enough signatures since this behavior is already present and should be improved for regular transaction (not loaded from a partially signed transaction JSON file) as well.
https://github.com/Concordium/concordium-client/issues/305

I added checks for the  `expiredTimestamp` in this PR.

Yes, we should remove the `signAndSubmit` command in my opinion as well:
https://github.com/Concordium/concordium-client/issues/306

I will add to the developer documentation your hack on creating a partially-signed transaction file without signing it with any keys. This can be useful if one tech-savvy person prepares the file but other people need to sign it.

PR #308, opened 2024-05-27
Remove command transactionSignAndSubmit

### LINKED ISSUE #306: Remove `signAndSubmit` command in `concordium-client`
Remove `signAndSubmit` command in `concordium-client`. 

This command was one of the first commands added to `concordium-client` and used to create any transaction type from a JSON file. Meanwhile `concordium-client` is more refined and has specific subcommands e.g. `stack run concordium-client -- contract update` to create (e.g. contract-update types of transactions) much more easily than loading the whole transaction information from a JSON file.
We recommend using the specific subcommands over the all-including `signAndSubmit` command.

In addition, the `signAndSubmit` command could be confused with the commands for `multi-sig` signing, so removing this command would be good to clean up the code base and avoid confusion.

DOBEN:
## Purpose

closes #306 

Under the subcommand `raw/legacy`, the sending of a transaction from a FILE given some keys for signing is still available.

## Changes

Remove command `transaction sign-and-submit`

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- Remove command `transaction sign-and-submit`.
 - Remove command `transaction send-shielded` to disable the transfer of CCD from the shielded balance of the account to the shielded balance of another account.
 - Remove command `account shield` to disable the transfer of CCD from the public balance to the shielded balance of an account.
 - Rename command `transaction submit` to `transaction sign-and-submit`.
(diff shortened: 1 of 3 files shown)


td202 (approved):
This is fine, but I think perhaps we should also remove the legacy `raw SendTransaction` command that does the same thing. Then we can remove a bunch of the code that supports these (`processTransaction`, `TransactionJSON`, `TransactionJSONPayload`, etc.).

eb-concordium (approved):

td202 (approved):

Issue #149, opened 2022-05-25
Time difference in cooldown period compared to account effective change time


mh-concordium:
**Bug Description**
While reducing the stake of a baker, I was observing cooldown time in the warning message and compared it to the time that is visible when inspecting account after transaction was finalized. There was around 1h difference between those.
 
**Steps to Reproduce**
- Reduce baker stake -> Observe time in the warning message (Figure1).
- Commit transaction -> tx finalized.
- Inspect the account that just had stake reduced -> Observe time at which stake will be reduced (Figure2).

**Expected Result**
I assume that the time should match.

**Actual Result**
Figure1
![image](https://user-images.githubusercontent.com/98821384/170236527-b67d2579-e108-494f-858d-6c9c364a25a3.png)

Figure2
![image](https://user-images.githubusercontent.com/98821384/170236633-b9d015ff-fc2f-464e-a1b5-2a62d2ad8b1a.png)

**Versions**
- Software Version:4.0.3

**Note:** Briefly discussed with Emil B and concluded that this issue could be fixed with https://github.com/Concordium/concordium-client/issues/142

┆Issue is synchronized with this [Jira Task](https://concordium.atlassian.net/browse/CB-1374) by [Unito](https://www.unito.io)

PR #315, opened 2024-08-20
Fix expiry time for pending stake changes

### LINKED ISSUE #142: Client returns cooldown without taking into account time to next payday
**Bug Description**
Time that is shown seems to be the cooldown time only.

**Steps to Reproduce**
Remove baker/delegator
Reduce baker/delegation stake

**Expected Result**
Time that is show should include time to next payday after the cooldown expires.

**Actual Result**
Only cooldown is taken into account and is shown as a feedback to the user.

**Versions**
- Software Version: concordium-client_4.0.3-0

td202:
## Purpose

Fixes #142 #149

This fixes the expiry time that is presented to the user for pending stake changes. This affects both when the user reduces the stake (and the expected expiry time is shown to them) and when the user views the state of an account with a pending change.

## Changes

- Revise `getBakerCooldown` and `getDelegatorCooldown` to start the cooldown from the next payday. (We assume that the cooldown duration is a multiple of the payday length for the calculation to be accurate, but we generally ensure this fact.)
- Introduce `correctPendingChange` to correct the pending change on an account (if applicable).
- Apply the correction when displaying account info.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/src/Concordium/Client/Runner.hs
+++ b/src/Concordium/Client/Runner.hs
@@ -107,6 +107,7 @@ import Data.Text (Text)
 import qualified Data.Text as Text
 import qualified Data.Text.Encoding as Text
 import Data.Time
+import qualified Data.Time.Clock as Clock
 import qualified Data.Tuple as Tuple
 import qualified Data.Vector as Vec
 import Data.Word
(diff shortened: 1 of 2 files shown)


src/Concordium/Client/Runner.hs
```diff
@@ -1269,6 +1295,61 @@ getCryptographicParametersOrDie bhInput = do
         StatusInvalid -> logFatal ["GRPC response contained an invalid status code."]
         RequestFailed err -> logFatal ["I/O error: " <> err]
 
+-- | Compute the time of the first payday after a given time.
+--  This is used for determining the time at which a cooldown will actually elapse.
```
Victor-N-Suadicani, line 1299: Just trying to understand some terminology here: What does "elapse" mean here? Is it different from "expire"? The fact that this takes the expiration time of the cooldown as input suggests it doesn't mean the same thing but I'm not sure what it means in that case. The variable is also called "cooldownEnd" - does a cooldown "end" when it "expires"?

src/Concordium/Client/Runner.hs
```diff
@@ -1269,6 +1295,61 @@ getCryptographicParametersOrDie bhInput = do
...
+    UTCTime ->
+    UTCTime
+firstPaydayAfter nextPayday epochDuration (Types.RewardPeriodLength ep) cooldownEnd =
+    if cooldownEnd <= nextPayday
+        then nextPayday
+        else
+            let timeDiff = Clock.diffUTCTime cooldownEnd nextPayday
+                paydayLength = Types.durationToNominalDiffTime (fromIntegral ep * epochDuration)
+                mult :: Word = ceiling (timeDiff / paydayLength)
+            in  Clock.addUTCTime (fromIntegral mult * paydayLength) nextPayday
```
Victor-N-Suadicani, line 1317: I'm not sure I fully understand this calculation. `paydayLength` sounds like it should be equal to the `ep` input variable which is described as "Length of a payday" but it seems like it is not equal. Could you maybe add a comment explaining why the calculation is done like this? Why is this the correct way to calculate it?

src/Concordium/Client/Runner.hs
```diff
@@ -1269,6 +1295,61 @@ getCryptographicParametersOrDie bhInput = do
...
+            <$> f asiPendingChange
+    pendingChange f Types.AccountStakingDelegated{..} =
+        ( \newPendingChange ->
+            Types.AccountStakingDelegated{asiDelegationPendingChange = newPendingChange, ..}
+        )
+            <$> f asiDelegationPendingChange
+    effectiveTime :: Traversal' (Types.StakePendingChange' t) t
+    effectiveTime _ Types.NoChange = pure Types.NoChange
+    effectiveTime f (Types.ReduceStake amt oldTime) = Types.ReduceStake amt <$> f oldTime
+    effectiveTime f (Types.RemoveStake oldTime) = Types.RemoveStake <$> f oldTime
```
Victor-N-Suadicani, line 1351: This function seems extremely information dense. It would be very helpful to have some comments that explain what is going on, what situation each `case` corresponds to and what the code is doing to address that situation and why.

As it is, I have a very hard time understanding what this function is doing (my Haskell may not be up to snuff in this case but that hopefully doesn't invalidate this point).

src/Concordium/Client/Runner.hs
```diff
@@ -3035,22 +3119,22 @@ processBakerConfigureCmd baseCfgDir verbose backend txOpts isBakerConfigure cbCa
...
+                    Right v -> getBakerCooldown v
             let removing = capital == 0
             if removing
                 then logWarn ["This will remove the validator."]
                 else logWarn ["The new staked value appears to be lower than the amount currently staked on chain by this validator."]
             let decreaseOrRemove = if removing then "Removing a validator" else "Decreasing the amount a validator is staking"
             logWarn [decreaseOrRemove ++ " will lock the stake of the validator for a cooldown period before the CCD are made available."]
             logWarn ["During this period it is not possible to update the validator's stake, or stop the validator."]
-            logWarn [[i|The current validator cooldown would last until approximately #{cooldownDate}|]]
+            logWarn [[i|The validator cooldown will last until approximately #{cooldownDate}|]]
```
Victor-N-Suadicani, line 3137: Out of curiosity, why do we say "approximately" here? Is it not a precise time? Why might it vary?

Victor-N-Suadicani (approved):
LGTM, but my Haskell skills are not up to par with some of this stuff - more comments could help and would probably be good regardless.

src/Concordium/Client/Runner.hs
td202: The actual timestamp of the payday block (which is when the amount actually leaves cooldown) is not precisely determined in advance. Typically, it will be the third block whose timestamp is after the (payday) epoch transition time.

Issue #318, opened 2024-08-28
Conflicting `--out` arguments for `validator add`


limemloh:
**Bug Description**
The `--out <file>` argument for the subcommand `validator add` is listed twice and does not allow for writing the validator credentials.

**Steps to Reproduce**

The issue can be seen when running `concordium-client validator add -h` as the list of arguments includes `--out FILE` twice: one for writing the transaction to file without submitting, another one for writing the resulting validator file.

![image](https://github.com/user-attachments/assets/919eb698-cd01-46a5-ab45-932224e1ce4b)

**Expected Result**
Allow the user to write the validator file.

**Actual Result**
Currently, providing the `--out FILE` argument results in the transaction being written to a file instead of being submitted.

**Versions**
- concordium-client: `7.0.0`
- OS: Linux

PR #321, opened 2024-09-05
fix: collision with argument `--out` for `./client validator add ...`

rasmus-kirk:
## Purpose

See #318 

## Checklist

      hard-to-understand areas.

### DIFF
--- a/src/Concordium/Client/Commands.hs
+++ b/src/Concordium/Client/Commands.hs
@@ -1739,7 +1739,7 @@ bakerAddCmd =
                         <*> blockCommission
                         <*> (option (eitherReader amountFractionFromStringInform) (long "delegation-finalization-commission" <> metavar "DECIMAL-FRACTION" <> help ("Fraction the validator takes in commission from delegators on finalization rewards. " ++ rangesHelpString "finalization reward commission")))
                     )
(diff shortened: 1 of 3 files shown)


src/Concordium/Client/Commands.hs
```diff
@@ -1739,7 +1739,7 @@ bakerAddCmd =
                         <*> blockCommission
                         <*> (option (eitherReader amountFractionFromStringInform) (long "delegation-finalization-commission" <> metavar "DECIMAL-FRACTION" <> help ("Fraction the validator takes in commission from delegators on finalization rewards. " ++ rangesHelpString "finalization reward commission")))
                     )
-                <*> optional (strOption (long "out" <> metavar "FILE" <> help "File to write the validator credentials to, in case of successful transaction. These can be used to start the node."))
+                <*> optional (strOption (long "validator-credentials-out" <> metavar "FILE" <> help "File to write the validator credentials to, in case of successful transaction. These can be used to start the node."))
```
soerenbf, line 1742: Why do we need this in the first place? Is it not just copying the input file to an output location if the transaction succeeds?

limemloh (approved):

PR #322, opened 2024-09-19
Report correct available balance with node version < 7.

td202:
## Purpose

For node version < 7, the available balance on an account is not returned in the `GetAccountInfo` query response. This change checks if that is the case and computes the correct balance, instead of simply returning 0.

The main purpose of this is to support the wallet proxy, allowing us to deploy it before node version 7 is deployed.

## Changes

- Revise the `FromProto` implementation for `AccountInfo` to calculate the available amount if it is not present in the protobuf.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/src/Concordium/Client/GRPC2.hs
+++ b/src/Concordium/Client/GRPC2.hs
@@ -729,7 +729,17 @@ instance FromProto Proto.AccountInfo where
                 Just asi -> fromProto asi
         aiAccountAddress <- fromProto $ ai ^. ProtoFields.address
         aiAccountCooldowns <- mapM fromProto $ ai ^. ProtoFields.cooldowns
-        aiAccountAvailableAmount <- fromProto $ ai ^. ProtoFields.availableBalance
+        -- The available balance is only provided by node version 7 onwards, so we compute it
+        -- if it is not present.
+        mAvailableBalance <- mapM fromProto $ ai ^. ProtoFields.maybe'availableBalance
+        let stakedTotal =
+                sum (cooldownAmount <$> aiAccountCooldowns) + case aiStakingInfo of
+                    AccountStakingDelegated{..} -> asiStakedAmount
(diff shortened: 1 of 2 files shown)


src/Concordium/Client/GRPC2.hs
```diff
@@ -729,7 +729,17 @@ instance FromProto Proto.AccountInfo where
                 Just asi -> fromProto asi
         aiAccountAddress <- fromProto $ ai ^. ProtoFields.address
         aiAccountCooldowns <- mapM fromProto $ ai ^. ProtoFields.cooldowns
-        aiAccountAvailableAmount <- fromProto $ ai ^. ProtoFields.availableBalance
+        -- The available balance is only provided by node version 7 onwards, so we compute it
+        -- if it is not present.
+        mAvailableBalance <- mapM fromProto $ ai ^. ProtoFields.maybe'availableBalance
+        let stakedTotal =
+                sum (cooldownAmount <$> aiAccountCooldowns) + case aiStakingInfo of
```
DOBEN, line 736: Are the `cooldowns` already included in the `stakedAmount` or not?
Comparing it with this implementation:

https://github.com/Concordium/concordium-net-sdk/blob/b8a1eb3f2a8a8bf418442f056d7bdf10b3d70e52/src/Types/AccountInfo.cs#L60C39-L60C54

drsk0 (approved):
Looks good, thanks!

src/Concordium/Client/GRPC2.hs
td202: No, they are not. Technically, it's probably fine not to include the cooldowns, because the available balance should only be missing for nodes prior to version 7, which means they don't support cooldowns anyway (and so the total cooldown amount will always be 0).

DOBEN (approved):

PR #349, opened 2025-04-08
Create protocol level token transaction

DOBEN:
## Purpose

https://linear.app/concordium/issue/CLI-149/implement-plt-creation-governance-transaction

```
stack run concordium-client -- consensus chain-update --help
```
```
stack run concordium-client -- consensus chain-update ./paylaod.json --key ./export_key.json 
```
## Changes

- Update base dependency
- Add support for creating PLT (protocol level tokens).


td202 (changes requested):
In [this PR](https://github.com/Concordium/concordium-base/pull/618) I've changed the JSON serialization for `CreatePLT` so that it can convert the parameter between JSON and CBOR representation. This should allow supplying the parameter inside the payload, as:
```json
{
  ...
  "payload" : {
     "updateType": "createPLT",
     "update": {
        "tokenSymbol": "PLT",
        ...
        "initializationParameters": {
           "name": "Protocol level token",
           "allowList": true, ...
         }
}
```
Also, I think we should create a new `plt` branch based on `main` to merge this to.

src/Concordium/Client/GRPC2.hs
```diff
@@ -1870,6 +1912,29 @@ instance FromProto Proto.ProtocolUpdate where
...
+        case deMkSerialize th of
+            Left err -> fromProtoFail $ "Unable to convert 'TokenModuleRef': " <> err
+            Right v -> return (TokenModuleRef v)
+
+instance FromProto ProtoPLT.CreatePLT where
+    type Output ProtoPLT.CreatePLT = CreatePLT
+    fromProto cpUpdate = do
+        _cpltTokenSymbol <- fromProto $ cpUpdate ^. PLTFields.tokenSymbol
+        _cpltGovernanceAccount <- fromProto $ cpUpdate ^. PLTFields.governanceAccount
+        let _cpltDecimals = fromIntegral $ cpUpdate ^. PLTFields.decimals
```
td202, line 1933: I think it would be sensible to check that the value is in range here.
```suggestion
        _cpltDecimals <- case toIntegralSized (cpUpdate ^. PLTFields.decimals) of
            Nothing -> fromProtoFail $ "CreatePLT: decimals out of range"
            Just converted -> return converted
```
(Requires `toIntegralSized` from `Data.Bits`.)

src/Concordium/Client/Runner.hs
```diff
@@ -2864,6 +2864,7 @@ processConsensusCmd action _baseCfgDir verbose backend =
                         Left err -> logFatal [fn ++ ": " ++ err]
                         Right r -> return r
             rawUpdate@Updates.RawUpdateInstruction{..} <- loadJSON rawUpdateFile
+
```
td202, line 2867:
```suggestion
```

src/Concordium/Client/Output.hs
```diff
@@ -889,6 +889,17 @@ showEvent verbose ciM = \case
         verboseOrNothing $ printf "baker %s with account %s suspended" (show bID) (show acc)
     Types.BakerResumed bID acc ->
         verboseOrNothing $ printf "baker %s with account %s resumed" (show bID) (show acc)
+    Types.TokenModuleEvent tokenEvent ->
+        let Types.TokenEventDetails bss = Types._teDetails tokenEvent
+            invalidCBOR = printf "Could not decode token event details of token module event with token id %s and type %s as valid CBOR. The hex value of the event details is %s." (show $ Types._teSymbol tokenEvent) (show $ Types._teType tokenEvent) (show $ Types._teDetails tokenEvent)
+            bsl = BSL.fromStrict $ BSS.fromShort bss
+            eventDetails = case deserialiseFromBytes decodeString bsl of -- Try to decode the event details as a CBOR string
```
td202, line 896: I don't think in general the event details will be a string. Possibly `decodeValue` (from `Codec.CBOR.JSON`) could be used here and then render the value as JSON.

td202 (approved):

PR #353, opened 2025-05-01
Add get token list

DOBEN:
## Purpose

```
stack run concordium-client -- raw GetTokenList --grpc-port 15001 --grpc-ip 0.0.0.0
```


td202 (approved):
Looks good!

PR #354, opened 2025-05-05
Add tokens to account info queries (COR-1357)

DOBEN:
## Purpose

```
stack run concordium-client -- raw GetAccountInfo 2 --grpc-port 15001 --grpc-ip 0.0.0.0

stack run concordium-client -- account show 2 --grpc-port 15001 --grpc-ip 0.0.0.0
```

Related to:
https://github.com/Concordium/concordium-base/pull/632

## Changes

- Subcommand `account show` now displays protocol level tokens info.
- Subcommand `raw GetAccountInfo` now displays protocol level tokens info.

Screenshot from 2025-05-07 15-16-16

Screenshot from 2025-05-07 15-15-55

### DIFF
--- a/src/Concordium/Client/Output.hs
+++ b/src/Concordium/Client/Output.hs
@@ -44,6 +44,7 @@ import qualified Concordium.Client.Types.ConsensusStatus as ConsensusStatus
 import Concordium.Client.Types.Contract.BuildInfo (showBuildInfo)
 import Concordium.Common.Time (DurationSeconds (durationSeconds))
 import Concordium.Types.Execution (Event' (ecEvents), SupplementedEvent)
+import qualified Concordium.Types.Queries.Tokens as Types
 import Control.Monad
 import Control.Monad.Writer
 import qualified Data.Aeson as AE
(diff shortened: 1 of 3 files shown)


drsk0 (approved):
Nice!

allanbrondum (approved):

src/Concordium/Client/Output.hs
```diff
@@ -195,20 +196,38 @@ showRevealedAttributes as =
...
+  where
+    indent = replicate 24 ' '
+    formatToken (Types.Token tid (Types.TokenAccountState (Types.TokenAmount digits decs) inAllowList inDenyList)) =
+        let amount = fromIntegral digits / (10 ^ decs :: Double)
+        in  unlines
+                [ indent ++ "Token ID:            " ++ show tid,
+                  indent ++ "Balance:             " ++ printf ("%." ++ show decs ++ "f") amount,
+                  indent ++ "In Allow List:       " ++ show inAllowList,
+                  indent ++ "In Deny List:        " ++ show inDenyList
+                ]
```
td202, line 210: I would suggest just having one line per PLT and having it something like
```
    50.000000 PLT
    0.100000  ETH
```
I would leave out the allow and deny list for now, as even the interface for them is subject to change.

td202 (approved):

src/Concordium/Client/Output.hs
DOBEN: I can make it one line but would like to keep the info about the `allow` and `deny` list since it is valid info.

DOBEN: Screenshot from 2025-05-08 12-55-42
To remove the quotes around "ETH" we can implement some custom show instead of deriving it at the types in another PR.
https://github.com/Concordium/concordium-base/blob/plt/haskell-src/Concordium/Types/Tokens.hs#L19 (edited)

PR #364, opened 2025-05-17
Add transaction plt mint/burn command (COR-1364)

DOBEN:
## Purpose

```
stack run concordium-client -- transaction plt mint --sender [your-sender-account]  --amount 1 --tokenId PLT
```
## Changes

- Add `transaction plt mint` command to mint protocol level tokens by the governance account.
- Add `transaction plt burn` command to burn protocol level tokens by the governance account.
- Refactored plt commands into one subcommand.

### DIFF
--- a/src/Concordium/Client/Commands.hs
+++ b/src/Concordium/Client/Commands.hs
@@ -224,11 +224,17 @@ data TransactionCmd
           trdTransactionOptions :: !(TransactionOpts (Maybe Energy))
         }
     | TransactionPLTTransfer
-        { tthReceiver :: !Text,
-          tthAmount :: !TokenAmount,
-          tthSymbol :: !Text,
-          tthMemo :: !(Maybe MemoInput),
-          tthOpts :: !(TransactionOpts (Maybe Energy))
+        { tptReceiver :: !Text,
+          tptAmount :: !TokenAmount,
+          tptSymbol :: !Text,
+          tptMemo :: !(Maybe MemoInput),
+          tptOpts :: !(TransactionOpts (Maybe Energy))
+        }
+    | TransactionPLTUpdateSupply
+        { tpusAction :: !Text,
+          tpusAmount :: !TokenAmount,
+          tpusSymbol :: !Text,
+          tpusOpts :: !(TransactionOpts (Maybe Energy))
         }
     deriving (Show)
 
@@ -721,6 +727,7 @@ transactionCmds =
                         <> transactionDeployCredentialCmd
                         <> transactionRegisterDataCmd
                         <> transactionPLTTransferCmd
+                        <> transactionPLTUpdateSupplyCmd
                     )
             )
             (progDesc "Commands for submitting and inspecting transactions.")
@@ -848,6 +855,20 @@ transactionPLTTransferCmd =
                 <*> transactionOptsParser
             )
             (progDesc "Transfer tokens from one account to another.")
+        )
+
+transactionPLTUpdateSupplyCmd :: Mod CommandFields TransactionCmd
+transactionPLTUpdateSupplyCmd =
+    command
+        "update-supply-plt"
+        ( info
+            ( TransactionPLTUpdateSupply
+                <$> strOption (long "action" <> metavar "ACTION" <> help "The action (either `mint` or `burn`)")
+                <*> option (eitherReader tokenAmountFromStringInform) (long "amount" <> metavar "TOKEN-AMOUNT" <> help "Amount of tokens to send.")
+                <*> strOption (long "tokenId" <> metavar "TOKEN_ID" <> help "Token id (Symbol) of the token.")
+                <*> transactionOptsParser
+            )
+            (progDesc "Mint or burn plt tokens.")
         )
 
 transactionWithScheduleCmd :: Mod CommandFields TransactionCmd
--- a/src/Concordium/Client/Commands.hs
+++ b/src/Concordium/Client/Commands.hs
@@ -13,6 +13,8 @@ module Concordium.Client.Commands (
     Interval (..),
     InteractionOpts (..),
     TransactionCmd (..),
+    PLTCmd (..),
+    TokenSupplyAction (..),
     AccountCmd (..),
     ModuleCmd (..),
     ContractCmd (..),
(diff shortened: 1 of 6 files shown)


src/Concordium/Client/Runner.hs
```diff
@@ -934,6 +934,33 @@ processTransactionCmd action baseCfgDir verbose backend =
...
+            baseCfg <- getBaseConfig baseCfgDir verbose
+            when verbose $ do
+                runPrinter $ printBaseConfig baseCfg
+                putStrLn ""
+
+            withClient backend $ do
+                tokenGovernanceOperation <- case actionText of
+                    "mint" -> pure $ CBOR.TokenMint amount
+                    "burn" -> pure $ CBOR.TokenBurn amount
+                    _ -> logFatal ["Only `mint` or `burn` supported for the `action` option."]
```
drsk0, line 950: Suggestion: "`Action` must be one of `mint`/`burn`."

drsk0 (approved):

src/Concordium/Client/Commands.hs
```diff
@@ -850,6 +857,20 @@ transactionPLTTransferCmd =
...
+        "update-supply-plt"
+        ( info
+            ( TransactionPLTUpdateSupply
+                <$> strOption (long "action" <> metavar "ACTION" <> help "The action (either `mint` or `burn`)")
+                <*> option (eitherReader tokenAmountFromStringInform) (long "amount" <> metavar "TOKEN-AMOUNT" <> help "Amount of tokens to send.")
+                <*> strOption (long "tokenId" <> metavar "TOKEN_ID" <> help "Token id (Symbol) of the token.")
+                <*> transactionOptsParser
+            )
+            (progDesc "Mint or burn plt tokens.")
+        )
```
td202, line 872: I think `concordium-client transaction update-supply-plt mint` is not a great user interface - it's long and difficult to remember. I'd suggest actually we gather all the PLT-related commands under a top level `plt` and have:
- `plt send` (currently `transaction transfer-plt`; `send` is congruent with the `transaction send` operation for sending CCDs).
- `plt mint` (rather than `transaction update-supply-plt mint`)
- `plt burn` (rather than `transaction update-supply-plt burn`)

src/Concordium/Client/Runner.hs
DOBEN: Changed it to:

````
`Action` option must be either `mint` or `burn`.
```

src/Concordium/Client/Runner.hs
```diff
@@ -897,46 +898,99 @@ processTransactionCmd action baseCfgDir verbose backend =
...
+    withClient backend $ do
+        tokenGovernanceOperation <- case tokenSupplyAction of
+            Mint -> pure $ CBOR.TokenMint amount
+            Burn -> pure $ CBOR.TokenBurn amount
+
+        let tokenGovernanceTransaction = CBOR.TokenGovernanceTransaction (Seq.fromList [tokenGovernanceOperation])
+        let bytes = CBOR.tokenGovernanceTransactionToBytes tokenGovernanceTransaction
+        let tokenParameter = Types.TokenParameter $ BS.toShort bytes
+
+        let symbol = tokenIdFromText symbolText
```
td202, line 983: `tokenIdFromText` needs to check that the symbol text is not longer than 255 encoded bytes, as otherwise it will break the invariant on `TokenId`, which will ultimately mean the encoded transaction is garbage.

src/Concordium/Client/Runner.hs
```diff
@@ -897,46 +898,99 @@ processTransactionCmd action baseCfgDir verbose backend =
...
+
+        let payload = Types.TokenGovernance symbol tokenParameter
+        let encodedPayload = Types.encodePayload payload
+
+        let nrgCost _ = return $ Just $ simpleTransferEnergyCost $ Types.payloadSize encodedPayload
```
td202, line 988: Probably shouldn't use this as it's tied to `simpleTransferCost`. Currently, `tokenHolderBaseCost` is used instead, though both happen to be 300.

src/Concordium/Client/Runner.hs
```diff
@@ -897,46 +898,99 @@ processTransactionCmd action baseCfgDir verbose backend =
...
-                let tokenHolderTransaction = CBOR.TokenHolderTransaction (Seq.fromList [tokenHolderTransfer])
-                let bytes = CBOR.tokenHolderTransactionToBytes tokenHolderTransaction
-                let tokenParameter = Types.TokenParameter $ BS.toShort bytes
+        let operation = CBOR.TokenTransferBuilder (Just amount) (Just tokenReceiver) maybeCborMemo
+        let eitherTokenTranfserBody = CBOR.buildTokenTransfer operation
+        tokenTransferBody <- case eitherTokenTranfserBody of
+            Right val -> return val
+            Left err -> logFatal ["Error creating token transfer body:", err]
+        let tokenHolderTransfer = CBOR.TokenHolderTransfer tokenTransferBody
+        let tokenHolderTransaction = CBOR.TokenHolderTransaction (Seq.fromList [tokenHolderTransfer])
```
td202, line 943:
```suggestion
        let tokenHolderTransaction = CBOR.TokenHolderTransaction (Seq.singleton tokenHolderTransfer)
```

src/Concordium/Client/Runner.hs
```diff
@@ -897,46 +898,99 @@ processTransactionCmd action baseCfgDir verbose backend =
...
+    when verbose $ do
+        runPrinter $ printBaseConfig baseCfg
+        putStrLn ""
+
+    withClient backend $ do
+        tokenGovernanceOperation <- case tokenSupplyAction of
+            Mint -> pure $ CBOR.TokenMint amount
+            Burn -> pure $ CBOR.TokenBurn amount
+
+        let tokenGovernanceTransaction = CBOR.TokenGovernanceTransaction (Seq.fromList [tokenGovernanceOperation])
```
td202, line 979:
```suggestion
        let tokenGovernanceTransaction = CBOR.TokenGovernanceTransaction (Seq.singleton tokenGovernanceOperation)
```

td202 (approved):
Looks good. A few small suggestions.

PR #372, opened 2025-06-04
COR-709: update operation costs

drsk0:
Update the energy cost calculation for PLT transactions to reflect the updated cost model. 

## Checklist

      hard-to-understand areas.

## CLA acceptance

_Remove if not applicable.

By submitting the contribution I accept the terms and conditions of the
Contributor License Agreement v1.0
- link: https://developers.concordium.com/CLAs/Contributor-License-Agreement-v1.0.pdf

- [x] I accept the above linked CLA.

### DIFF
--- a/src/Concordium/Client/Output.hs
+++ b/src/Concordium/Client/Output.hs
@@ -921,7 +921,7 @@ showEvent verbose ciM = \case
                     bsl = BSL.fromStrict $ BSS.fromShort bss
 
                     -- First decoding attempt using the token event decoder for the add/remove to/from allow/deny list events.
-                    decodedTokenEvent = Cbor.decodeTokenEvent (Cbor.EncodedTokenEvent (Types.tokenEventTypeBytes etmeType) (Just bss))
+                    decodedTokenEvent = Cbor.decodeTokenEvent (Cbor.EncodedTokenEvent etmeType etmeDetails)
 
                     decodedCustomTokenEvent = case decodedTokenEvent of
                         Right (Cbor.AddAllowListEvent holder) ->
--- a/src/Concordium/Client/Runner.hs
+++ b/src/Concordium/Client/Runner.hs
@@ -1038,7 +1038,7 @@ handlePLTModifyList backend baseCfgDir verbose modifyListAction account tokenIdT
         let payload = Types.TokenGovernance tokenId tokenParameter
         let encodedPayload = Types.encodePayload payload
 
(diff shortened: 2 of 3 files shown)


src/Concordium/Client/Output.hs
```diff
@@ -921,7 +921,7 @@ showEvent verbose ciM = \case
                     bsl = BSL.fromStrict $ BSS.fromShort bss
 
                     -- First decoding attempt using the token event decoder for the add/remove to/from allow/deny list events.
-                    decodedTokenEvent = Cbor.decodeTokenEvent (Cbor.EncodedTokenEvent (Types.tokenEventTypeBytes etmeType) (Just bss))
+                    decodedTokenEvent = Cbor.decodeTokenEvent (Cbor.EncodedTokenEvent etmeType etmeDetails)
```
DOBEN, line 924: please update the changelog.

src/Concordium/Client/Runner.hs
```diff
@@ -1034,7 +1038,7 @@ handlePLTModifyList backend baseCfgDir verbose modifyListAction account tokenIdT
         let payload = Types.TokenGovernance tokenId tokenParameter
         let encodedPayload = Types.encodePayload payload
 
-        let nrgCost _ = return $ Just $ tokenGovernanceTransactionEnergyCost $ Types.payloadSize encodedPayload
+        let nrgCost _ = return $ Just $ tokenGovernanceTransactionEnergyCost (Types.payloadSize encodedPayload) (Cost.tokenListOperationCost)
```
DOBEN, line 1041:
```suggestion
        let nrgCost _ = return $ Just $ tokenGovernanceTransactionEnergyCost (Types.payloadSize encodedPayload) Cost.tokenListOperationCost
```

DOBEN (approved):

src/Concordium/Client/Output.hs
DOBEN: In general, the changelog should be updated (not particularly related to the above line). This line was just the first line in the review where I could add a comment.

E.g. aligning the `node` release and the `concordium-client` release to ensure that both have the same feature set when we make the cut.

drsk0: I think we should only add functionally important changes to the changelog. The changelog should be a place to easily figure out, how the software changed from release to release. If you just mirror the git history, you'll end up with a hard to read and redundant document. I don't think this change falls into the functionally important category.

DOBEN: If a transaction costs more or less in a new release, it's important to be aware of. However, as it is only an alpha release, feel free to proceed without a changelog entry as we are internally aware of the change and can align the `node`/`concordium-client` release.

PR #382, opened 2025-06-18
Extendable plt events

DOBEN:
## Purpose

- Propagate extendable plt event changes.

### DIFF
--- a/src/Concordium/Client/GRPC2.hs
+++ b/src/Concordium/Client/GRPC2.hs
@@ -87,7 +87,6 @@ import qualified Data.Text.Encoding as TE
 
 import Concordium.Client.Types.ConsensusStatus
 import qualified Concordium.Crypto.BlockSignature as BlockSignature
-import qualified Concordium.Types.ProtocolLevelTokens.CBOR as Cbor
 import qualified Proto.V2.Concordium.Kernel as ProtoKernel
 import qualified Proto.V2.Concordium.ProtocolLevelTokens as ProtoPLT
 import qualified Proto.V2.Concordium.ProtocolLevelTokens_Fields as ProtoFieldsPLT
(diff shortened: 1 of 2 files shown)


allanbrondum (approved):

drsk0 (approved):

src/Concordium/Client/GRPC2.hs
```diff
@@ -2168,19 +2167,19 @@ protoToTokenEvent event = do
...
+            return $ TokenBurn tokenId target amount
         ProtoPLT.TokenEvent'TransferEvent e -> do
-            fromAccount <- Cbor.holderAccountAddress <$> fromProto (e ^. ProtoFieldsPLT.from)
-            toAccount <- Cbor.holderAccountAddress <$> fromProto (e ^. ProtoFieldsPLT.to)
+            fromAddress <- fromProto $ e ^. ProtoFieldsPLT.from
+            toAddress <- fromProto $ e ^. ProtoFieldsPLT.to
             amount <- fromProto $ e ^. ProtoFieldsPLT.amount
             memo <- fromProtoMaybe $ e ^. ProtoFields.maybe'memo
-            return $ TokenTransfer tokenId fromAccount toAccount amount memo
+            return $ TokenTransfer tokenId fromAddress toAddress amount memo
```
td202, line 2182: I think it's a good idea to use record syntax, especially when constructors have many parameters and where there are multiple parameters of the same type (e.g. the from and to addresses). This makes it easier to catch any issues that might come from any changes.
```suggestion
            return TokenTransfer {
                    ettTokenId = tokenId,
                    ettFrom = fromAddress,
                    ettTo = toAddress,
                    ettAmount = amount,
                    ettMemo = memo
                }
```
(Typically, I would name the variables the same as the fields and use record wildcards with `TokenTransfer {..}`.)

PR #386, opened 2025-07-04
basepackage

allanbrondum:
## Purpose

Use base as local package and not as dependency. This mean Haskell language server works with base.

## Changes

_Describe the changes that were needed.

## Checklist

      hard-to-understand areas.


.github/workflows/ci.yaml
```diff
@@ -100,4 +100,4 @@ jobs:
       - name: Stack build # build all the binaries and tests, but don't run the tests
         run: stack build --test --no-run-tests --ghc-options "-Werror"
       - name: Stack test # run the tests
-        run: stack build --test
+        run: stack build --test concordium-client
```
allanbrondum, line 103: restrict such that we don't run base tests now that base is a package in our workspace

DOBEN (approved):
Nice. Jumping around between types from base and concordium-client works now with the Haskell Language Server.

PR #387, opened 2025-07-08
Handle renormalizing amounts to the correct number of decimals.

td202:
## Purpose

Closes [COR-1457](https://linear.app/concordium/issue/COR-1457/concordium-client-handle-plt-amounts)

Convert PLT amounts to the correct number of decimals, querying the node.

## Changes

- Add 'renormalizeTokenAmount' helper function
- Query the token info and renormalize token amounts in plt transfer, mint and burn operations.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/src/Concordium/Client/Runner.hs
+++ b/src/Concordium/Client/Runner.hs
@@ -69,6 +69,7 @@ import Concordium.Types.HashableTo
 import qualified Concordium.Types.InvokeContract as InvokeContract
 import Concordium.Types.Parameters
 import qualified Concordium.Types.Queries as Queries
+import Concordium.Types.Tokens
 import qualified Concordium.Types.Transactions as Types
 import qualified Concordium.Types.Updates as Updates
 import qualified Concordium.Utils.Encryption as Password
@@ -81,7 +82,7 @@ import Codec.CBOR.JSON
 import Codec.CBOR.Write
 import Concordium.Client.Types.Contract.BuildInfo (extractBuildInfo)
 import qualified Concordium.Types.ProtocolLevelTokens.CBOR as CBOR
-import Concordium.Types.Queries.Tokens (TokenAmount)
+import Concordium.Types.Queries.Tokens (TokenInfo (..), TokenState (..))
 import Control.Arrow (Arrow (second))
 import Control.Concurrent (threadDelay)
 import Control.Exception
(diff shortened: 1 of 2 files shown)


src/Concordium/Client/Runner.hs
```diff
@@ -907,6 +908,23 @@ processTransactionCmd action baseCfgDir verbose backend =
                 TransactionPLTModifyList modifyListAction account tokenId txOpts ->
                     handlePLTModifyList backend baseCfgDir verbose modifyListAction account tokenId txOpts
 
+-- | Renormalize a 'TokenAmount' to conform to the number of decimal places expected by the
+--  token. If more than the expected number of decimals are given, this fails with an error.
+--  Similarly, if the resulting value would exceed the maximum representatable amount of the token,
+--  this fails with an error.
+renormalizeTokenAmount :: (MonadIO m) => TokenInfo -> TokenAmount -> m TokenAmount
```
allanbrondum, line 915: Would it be better to let the commands parse to some (exact) decimal floating point type and then convert from that type to `TokenAmount` here. It will likely mean that we don't have to implement our own floating point logic, neither in `tokenAmountFromString`, nor this function. We can use the decimal type to parse and rescale.

src/Concordium/Client/Runner.hs
```diff
@@ -907,6 +908,23 @@ processTransactionCmd action baseCfgDir verbose backend =
...
+    | amountDecimals > expectDecimals =
+        logFatal [[i|Token amount #{tokenAmountToString amount} has #{amountDecimals} decimals, but #{tiTokenId ti} only supports #{expectDecimals}.|]]
+    | amountDecimals == expectDecimals = return amount
+    | preNorm > fromIntegral (maxBound :: TokenRawAmount) =
+        logFatal [[i|Token amount #{tokenAmountToString amount} exceeds the maximum representable token amount for #{tiTokenId ti}.|]]
+    | otherwise = return TokenAmount{taValue = fromIntegral preNorm, taDecimals = expectDecimals}
+  where
+    amountDecimals = taDecimals amount
+    expectDecimals = tsDecimals (tiTokenState ti)
+    preNorm = toInteger (taValue amount) * 10 ^ (expectDecimals - amountDecimals)
```
allanbrondum, line 926: Question (I don't know): what happens at overflow?

src/Concordium/Client/Runner.hs
allanbrondum: (also, it means `TokenAmount` is never created at the wrong scale)

allanbrondum: Alternatively, if we implement our own, we should probably have test of the functions.

eb-concordium (approved):

